PR #16, opened 2019-09-10
Describe BIP32 Paths

keepkeyjon:


packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -58,6 +60,78 @@ export function isKeepKey(wallet: any): wallet is KeepKeyHDWallet {
...
+  }[scriptType]
+
+  let index = path[2] & 0x7fffffff
+
+  if (wholeAccount) {
+    return `${coin} ${script}Account #${index}`
+  } else {
+    let change = path[3] == 1 ? 'Change ' : ''
+    let address = path[4]
+    return `${coin} ${script}Account #${index}, ${change}Address #${address}`
```
keepkeyjon: @ascension I'm considering returning an object from these, something like:

```
return {
  verbose: `${coin} ${script}Account #${index}, ${change}Address #${address}`,
  accountIdx: index,
  addressIdx: address,
  change: path[3] === 1
}
```

ascension: Can we add `coin` as a property.

packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -58,6 +60,78 @@ export function isKeepKey(wallet: any): wallet is KeepKeyHDWallet {
...
+  if (path[4] != 0)
+    return undefined
+
+  let index = path[2] & 0x7fffffff
+  return `Ethereum Account #${index}`
+}
+
+function describeUTXOPath (path: BIP32Path, coin: Coin, scriptType: BTCInputScriptType) {
+  if (!this.btcSupportsCoin(coin))
+    return undefined
```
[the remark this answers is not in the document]
keepkeyjon: It's not that the path is invalid in those cases... every path is valid. The difference is that some paths are used by the device to store coins, and others are not. Giving feedback severely complicates this function, and not in a good way.

> Another thing to consider would be giving the user a valid until this point in the path?

Don't think that makes sense.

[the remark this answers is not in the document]
keepkeyjon: Maybe a better thing to return in those cases is `''`. I'd strongly prefer not to throw.

.circleci/config.yml
```diff
@@ -30,7 +31,7 @@ orbs:
                 - .yarn-cache
           - save_cache:
               name: Save Yarn Package Cache
-              key: yarn-packages-{{ checksum "yarn.lock" }}
+              key: cache-v4-{{ checksum "yarn.lock" }}
```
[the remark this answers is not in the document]
keepkeyjon: Yes, which I fixed in one of the patches in this series. Needed a cache bump. The bug was that the `/dist` folders were getting put in the cache, which means that the tests are testing against the wrong version of the package.

packages/hdwallet-keepkey/src/keepkey.ts
keepkeyjon: so just the value that got passed in? seems not too useful

packages/hdwallet-trezor/src/trezor.ts
```diff
@@ -44,6 +49,115 @@ export function isTrezor(wallet: HDWallet): wallet is TrezorHDWallet {
...
+
+  if (path.length !== 3 && path.length !== 5)
+    return unknown
+
+  if ((path[0] & 0x80000000) >>> 0 !== 0x80000000)
+    return unknown
+
+  let purpose = path[0] & 0x7fffffff
+
+  if (![44, 49, 84].includes(purpose))
```
elmutt: Do we have these purposes defined somewhere like an enum?

keepkeyjon: They're self-documenting.

https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki
https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki

PR #19, opened 2019-09-11
Remove calls to CQv1 in sandbox, use static JSON files instead

millsmcilroy:


examples/sandbox/index.ts
```diff
@@ -435,17 +441,14 @@ $btcTx.on('click', async (e) => {
     const response = await fetch(lookup)
     const data = await response.json()
     const hex = data[0].hex
-    const txLookup = 'https://btc.coinquery.com/api/tx/'+txid
```
keepkeyjon: Mind doing the same for this bit of each of them too?
```
const lookup = `https://api.ledgerwallet.com/blockchain/v2/btc/transactions/${txid}/hex`
    const response = await fetch(lookup)
    const data = await response.json()
```

PR #25, opened 2019-09-20
Add Portis

elmutt:
- Add portis package containing `adapter.ts` and `portis.ts`
- Add portis test suite to integration


examples/sandbox/index.ts
```diff
@@ -28,10 +30,14 @@ import * as dashTxJson from './json/dashTx.json'
 import * as dogeTxJson from './json/dogeTx.json'
 import * as ltcTxJson from './json/ltcTx.json'
 
+import Portis from "@portis/web3";
+
 const keyring = new Keyring()
 
+const portis = new Portis('ff763d3d-9e34-45a1-81d1-caa39b9c64f9', 'mainnet');
```
keepkeyjon: This should happen inside the `Web3PortisAdapter`. If you need to pass arguments to it, follow the pattern that the `TrezorConnect` one uses.

examples/sandbox/index.ts
```diff
@@ -28,10 +30,14 @@ import * as dashTxJson from './json/dashTx.json'
 import * as dogeTxJson from './json/dogeTx.json'
 import * as ltcTxJson from './json/ltcTx.json'
 
+import Portis from "@portis/web3";
```
keepkeyjon: This should also be inside the adapter.

examples/sandbox/index.ts
```diff
@@ -15,6 +15,8 @@ import { WebUSBKeepKeyAdapter } from '@shapeshiftoss/hdwallet-keepkey-webusb'
 import { TCPKeepKeyAdapter } from '@shapeshiftoss/hdwallet-keepkey-tcp'
 import { TrezorAdapter } from '@shapeshiftoss/hdwallet-trezor-connect'
 import { WebUSBLedgerAdapter } from '@shapeshiftoss/hdwallet-ledger-webusb'
+import { Web3PortisAdapter } from '@shapeshiftoss/hdwallet-web3-portis'
```
ascension: small nit `hdwallet-portis-web3`

packages/hdwallet-web3-portis/src/adapter.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+  
+    public static useKeyring (keyring: Keyring) {
+      console.log('portis adapter useKeyring')
+      return new Web3PortisAdapter(keyring)
+    }
+  
+    public async initialize (portis: any): Promise<number> {
+      console.log('portis adapter initialize')
+      let transport = new PortisTransport(this.keyring)
```
keepkeyjon: If you're going to have a transport, you should put the `portis` object in the transport. Otherwise it's an unnecessary abstraction.

packages/hdwallet-web3-portis/src/adapter.ts
```diff
@@ -0,0 +1,36 @@
...
+      let wallet = new Web3PortisHDWallet(portis, transport)
+      await wallet.initialize()
+      this.keyring.add(wallet)
+      return Object.keys(this.keyring.wallets).length
+    }
+
+    public async pairDevice (deviceId: string): Promise<HDWallet> {
+      console.log('portis adapter pairDevices', deviceId)
+      const wallet = this.keyring.get(deviceId)
```
keepkeyjon: What did we decide `deviceId` meant for Portis wallets?

packages/hdwallet-web3-portis/src/portisTransport.ts
```diff
@@ -0,0 +1,51 @@
...
+    public keyring: Keyring
+  
+    constructor (keyring: Keyring) {
+      console.log('PortisTransport constructor')
+      super(keyring)
+    }
+  
+    public getDeviceID(): string {
+      console.log('portisTransport getDeviceID()')
+      return 'portisTransport getDeviceId()'
```
keepkeyjon: oh. uh.....

packages/hdwallet-web3-portis/src/portisTransport.ts
```diff
@@ -0,0 +1,51 @@
+import {
+    Keyring,
+    Transport
+  } from '@shapeshiftoss/hdwallet-core'
```
keepkeyjon: fucky indentation.

packages/hdwallet-web3-portis/src/web3Portis.ts
```diff
@@ -0,0 +1,214 @@
...
+    ETHWallet,
+    ETHGetAddress,
+    ETHSignTx,
+    ETHGetAccountPath,
+    ETHAccountPath,
+    ETHSignMessage,
+    ETHSignedMessage,
+    ETHVerifyMessage,
+    ETHSignedTx
+  } from "@shapeshiftoss/hdwallet-core";
```
keepkeyjon: more bad indentation

examples/sandbox/index.ts
ascension: This should be part of the adapter similar to how Trezor is implemented.

packages/hdwallet-web3-portis/src/web3Portis.ts
```diff
@@ -0,0 +1,214 @@
...
+      throw new Error("Method not implemented.");
+    }
+    loadDevice(msg: LoadDevice): Promise<void> {
+      console.log('Web3PortisHDWallet loadDevice')
+      throw new Error("Method not implemented.");
+    }
+  
+    public async ethSupportsNetwork (chain_id: number): Promise<boolean> {
+      console.log('Web3PortisHDWallet ethSupportsNetwork')
+      return true
```
keepkeyjon: may want to make this `chain_id === 1` until they support more ETH networks.

examples/sandbox/index.ts
keepkeyjon: should be two packages: `hdwallet-web3` and `hdwallet-portis`. And I don't think we need a transport for it, since there's only ever going to be one way to talk to a Portis wallet.

packages/hdwallet-web3-portis/src/web3Portis.ts
ascension: @keepkeyjon I think we need to add `prettier` as a pre-commit hook 😂

keepkeyjon: 😐

packages/hdwallet-web3-portis/src/adapter.ts
KeepKeyBrett: @keepkeyjon as of now, wallets are identified by their corresponding eth address, but idk if that's the best long term solution. The ETH address for the wallet is what gets passed in every time the `portis.onActiveWalletChange` callback gets invoked.

majorhayes:
This should be a DRAFT PR if it isn't supposed to be merged.  When creating a PR, there is a drop down menu on the create button that let's you select DRAFT.

keepkeyjon:
> This should be a DRAFT PR if it isn't supposed to be merged. When creating a PR, there is a drop down menu on the create button that let's you select DRAFT.

Too bad you can't make a PR draft after the fact :/

examples/sandbox/index.ts
elmutt: Its created here to pass it thru to HD wallet

ascension: @elmutt The point of the adapter is to encapsulate all this logic. see @keepkeyjon comments above

```
      const portisAdapter = PortisAdapter.useKeyring(keyring, config)
```

elmutt: This is the sandbox

elmutt: Yep.  thats a different line.  im doing something similar with:

```
const portisAdapter = PortisAdapter.useKeyring(keyring, config)
```

elmutt: ok i . see what u mean.  was confused what line u meant

packages/hdwallet-portis/src/portis.ts
```diff
@@ -223,18 +223,16 @@ import {
       return Promise.resolve()
     }
 
+    // TODO this needs to handle more than just eth
     public async getPublicKeys(msg: GetPublicKey[]): Promise<PublicKey[]> {
       console.log('GET PUBLIC KEYS')
-
-      const portisResult = await this.portis.getExtendedPublicKey("m/44'/60'/0'");
+      const portisResult = await this.portis.getExtendedPublicKey("m/44'/60'/0'")
```
keepkeyjon: We should probably ask Portis folks for an API that gives more than one at a time.

keepkeyjon: That way if they want to add some permission dialog, we don't have to bug the user over and over.

.gitignore
```diff
@@ -14,3 +14,4 @@ dist
 .rpt2_*
 .cache
 .npmrc
+junit.xml
```
millsmcilroy: :pray:

examples/sandbox/index.ts
```diff
@@ -96,6 +104,22 @@ $ledger.on('click', async (e) => {
...
+$portis.on('click',  async (e) => {
+  e.preventDefault()
+  wallet = await portisAdapter.pairDevice()
+  window['wallet'] = wallet
+
+  let deviceId = 'nothing'
+  try {
+    deviceId  = await wallet.getDeviceID()
+  } catch( e ) {
+    console.log('ERROR', { e })
```
millsmcilroy: console.error() for errors

examples/sandbox/index.ts
```diff
@@ -96,6 +104,22 @@ $ledger.on('click', async (e) => {
...
+  }
+  console.log('portis deviceId IS ', deviceId)
```
millsmcilroy: Do we need this log?

examples/sandbox/index.ts
```diff
@@ -288,35 +318,39 @@ $getXpubs.on('click', (e) => {
   e.preventDefault()
   if (!wallet) { $manageResults.val("No wallet?"); return}
   // Get Ethereum path
-  const { hardenedPath, relPath } = wallet.ethGetAccountPaths({coin: "Ethereum", accountIdx: 0})[0]
+  const { hardenedPath } = wallet.ethGetAccountPaths({coin: "Ethereum", accountIdx: 0})[0]
```
millsmcilroy: Good call

examples/sandbox/index.ts
keepkeyjon:
```suggestion
```

packages/hdwallet-portis/src/portis.ts
```diff
@@ -0,0 +1,386 @@
...
+    return Promise.resolve({ msg: msg.msg })
+  }
+
+  public sendPin(pin: string): Promise<void> {
+    // no concept of pin in Portis
+    return Promise.resolve()
+  }
+
+  public sendPassphrase(passphrase: string): Promise<void> {
+    // cannot send passphrase to Portis. Could show the widget?
```
keepkeyjon: reasonable

packages/hdwallet-portis/src/portis.ts
```diff
@@ -0,0 +1,386 @@
...
+    }
+  }
+
+  public ethNextAccountPath (msg: ETHAccountPath): ETHAccountPath | undefined {
+    // Portis only supports one account for eth
+    return undefined
+  }
+
+  public async isInitialized (): Promise<boolean> {
+    return false
```
keepkeyjon: This is asking "does the wallet have a seed in it". Better to assume it does when we have no way of checking.
```suggestion
    return true
```

packages/hdwallet-portis/src/portis.ts
```diff
@@ -0,0 +1,386 @@
...
+    // Portis only supports one account for eth
+    return undefined
+  }
+}
+
+export function info () {
+  return new PortisHDWalletInfo()
+}
+
+export function create (portis): PortisHDWallet {
```
millsmcilroy: Is it possible to declare the portis param as type Portis? `function create (portis: Portis): PortisHDWallet`

packages/hdwallet-portis/src/portis.ts
```diff
@@ -0,0 +1,386 @@
...
+  public async getPublicKeys(msg: GetPublicKey[]): Promise<PublicKey[]> {
+    const publicKeys = []
+    this.portisCallInProgress = new Promise( async (resolve, reject) => {
+      await this.portisCallInProgress
+      for (let i = 0; i < msg.length; i++) {
+        const { addressNList } = msg[i];
+        const portisResult = await this.portis.getExtendedPublicKey(addressNListToBIP32(addressNList))
+        const { result, error } = portisResult
+        if(error) {
+          console.log('getPublicKeys error', error)
```
keepkeyjon: should throw something here. the caller assumes the input list, and output list have the same length, and will be misled / confused by this.

packages/hdwallet-portis/src/portis.ts
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  public disconnect (): Promise<void> {
+    return Promise.resolve()
+  }
+
+  public async getPublicKeys(msg: GetPublicKey[]): Promise<PublicKey[]> {
+    const publicKeys = []
+    this.portisCallInProgress = new Promise( async (resolve, reject) => {
+      await this.portisCallInProgress
```
keepkeyjon:
```suggestion
      try {
          await this.portisCallInProgress
      } catch (e) {
          console.error(e)
      }
```

packages/hdwallet-portis/src/portis.ts
```diff
@@ -0,0 +1,386 @@
...
+    return [{
+      addressNList: [ 0x80000000 + 44, 0x80000000 + slip44ByCoin(msg.coin), 0x80000000 + msg.accountIdx, 0, 0 ],
+      hardenedPath: [ 0x80000000 + 44, 0x80000000 + slip44ByCoin(msg.coin), 0x80000000 + msg.accountIdx ],
+      relPath: [ 0, 0 ],
+      description: "Portis"
+    }]
+  }
+
+  public async hasOnDevicePinEntry (): Promise<boolean> {
+    return false
```
keepkeyjon: This feature test means "should I expect to get & handle pin requests in the client UI? or does the device handle that for us?". Since Portis doesn't have PINs, the client doesn't have to deal with those, hence this should return `true`:

```suggestion
    return true
```

examples/sandbox/index.ts
```diff
@@ -96,6 +104,21 @@ $ledger.on('click', async (e) => {
...
+$portis.on('click',  async (e) => {
+  e.preventDefault()
+  wallet = await portisAdapter.pairDevice()
+  window['wallet'] = wallet
+
+  let deviceId = 'nothing'
+  try {
+    deviceId  = await wallet.getDeviceID()
+  } catch( e ) {
+    console.error('ERROR', { e })
```
keepkeyjon:
```suggestion
    console.error(e)
```

PR #29, opened 2019-09-23
feat: handle Ethereum in ledger.getPublicKeys

millsmcilroy:


packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -328,13 +336,15 @@ export class LedgerHDWallet implements HDWallet, BTCWallet, ETHWallet {
       const coinType: number = parseInt(bip32path.split("/")[1], 10)
       const account: number = parseInt(bip32path.split("/")[2], 10)
       const childNum: number = (0x80000000 | account) >>> 0
+      const coin = networksUtil[coinType] || networksUtil[0] // use Bitcoin's # for Ethereum
```
keepkeyjon: I think we should only do this for specific coinType values (aka `60` in this case), that way we're not providing a default for everything. Recommend doing a switch, since we'll probably want to add more of them later.

millsmcilroy: I added Ethereum to networksUtil

keepkeyjon: ooh, even better

PR #30, opened 2019-09-24
fix: TransactionBuilder requires output.amount to be a number

millsmcilroy:
BTCSignTxOutput.amount is declared as type number but for whatever reason that type is not being enforced and strings are able to be passed in for amount. TransactionBuilder.addOutput() throws if it is passed a string for amount so I added a cast to number to fix this.


keepkeyjon:
mind adding the same check to KK and Trezor also, please?

PR #32, opened 2019-09-25
FEAT: get ledger firmware version

dviramontes:
changes:
- gets ledger firmware version from @ledger-live-common lib
- adds a new `yarn test:ci` target for generating coverage only on CI

### DIFF
--- a/packages/hdwallet-core/src/wallet.ts
+++ b/packages/hdwallet-core/src/wallet.ts
@@ -94,7 +94,8 @@ export interface PathDescription {
   accountIdx?: number,
   addressIdx?: number,
   isChange?: boolean,
-  wholeAccount?: boolean
+  wholeAccount?: boolean,
+  isPrefork?: boolean
 }
 
 type CoinWallets = BTCWallet | ETHWallet | DebugLinkWallet
--- a/packages/hdwallet-core/src/wallet.ts
+++ b/packages/hdwallet-core/src/wallet.ts
@@ -204,6 +204,11 @@ export interface HDWallet extends HDWalletInfo {
   getDeviceID (): Promise<string>
 
   /**
+   * Retrieve the wallet's info
+   */
+  getDeviceInfo (): Promise<any>
+
+  /**
    * Retrieve the name of the model of wallet, eg 'KeepKey' or 'Trezor One'
    */
   getModel (): Promise<string>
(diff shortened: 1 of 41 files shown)


packages/hdwallet-core/src/wallet.ts
```diff
@@ -203,6 +203,11 @@ export interface HDWallet extends HDWalletInfo {
    */
   getDeviceID (): Promise<string>
 
+  /**
+   * Retrieve the wallet's info
+   */
+  getDeviceInfo (): Promise<any>
```
keepkeyjon: It would be better to put the new functionality here under the existing `getFeatures()` call, rather than adding new API surface.

dviramontes: gotcha, I'll try that

packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -284,6 +284,10 @@ export class KeepKeyHDWallet implements HDWallet, BTCWallet, ETHWallet, DebugLin
     return (await this.getFeatures()).model;
   }
 
+  public async getFirmwareVersion(): Promise<string> {
+    return ""
```
keepkeyjon: should pull this out of the features object:

`${features.majorVersion}.${features.minorVersion}.${features.patchVersion}`

packages/hdwallet-trezor/src/trezor.ts
```diff
@@ -275,6 +275,10 @@ export class TrezorHDWallet implements HDWallet, BTCWallet, ETHWallet {
     return features.device_id
   }
 
+  public async getFirmwareVersion (): Promise<string> {
+    return ""
```
keepkeyjon: likewise, except the fields are named differently, IIRC

packages/hdwallet-ledger-u2f/src/transport.ts
```diff
@@ -2,6 +2,7 @@ import { Coin, makeEvent, Keyring } from '@shapeshiftoss/hdwallet-core'
 import { LedgerTransport } from '@shapeshiftoss/hdwallet-ledger'
 import Eth from '@ledgerhq/hw-app-eth'
 import Btc from '@ledgerhq/hw-app-btc'
+import getDeviceInfo from '@ledgerhq/live-common/lib/hw/getDeviceInfo'
```
keepkeyjon: probably also want this in `hdwallet-ledger-webusb` as well

PR #34, opened 2019-09-25
create portis object during pair and make initialize noop

elmutt:
create portis object during pair and make initialize noop

### DIFF
--- a/packages/hdwallet-portis/src/adapter.ts
+++ b/packages/hdwallet-portis/src/adapter.ts
@@ -5,8 +5,6 @@ import {
 
 import { PortisHDWallet } from './portis'
 import Portis from "@portis/web3";
-import Web3 from 'web3'
-import { getEthAddress } from './utils'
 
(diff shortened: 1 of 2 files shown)


packages/hdwallet-portis/src/adapter.ts
```diff
@@ -5,40 +5,35 @@ import {
...
-import { getEthAddress } from './utils'
 
 type PortisWallet = any
 
 export class PortisAdapter {
     keyring: Keyring
     portis: any
   
-    private constructor (keyring: Keyring, args: { portis?: PortisWallet, portisAppId?: string }) {
+    private constructor (keyring: Keyring) {
```
keepkeyjon: I think we should still pass the `AppId` in via `useKeyring()`, as that will be a more natural place to provide it from WalletContext than to have to pass it to `pairDevice()`.

elmutt: done

PR #36, opened 2019-09-25
handle portis wallet changed

elmutt:
handle portis wallet changed.  change spacing

### DIFF
--- a/packages/hdwallet-portis/src/adapter.ts
+++ b/packages/hdwallet-portis/src/adapter.ts
@@ -1,6 +1,7 @@
 import {
   Keyring,
-  HDWallet
+  HDWallet,
+  Events
 } from '@shapeshiftoss/hdwallet-core'
 
 import { PortisHDWallet } from './portis'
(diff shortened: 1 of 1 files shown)


packages/hdwallet-portis/src/adapter.ts
```diff
@@ -9,34 +10,43 @@ import Portis from "@portis/web3";
...
-      let wallet = new PortisHDWallet(this.portis)
-      await wallet.initialize()
-      const deviceId = await wallet.getDeviceID()
-      this.keyring.add(wallet, deviceId)
-      return wallet
-    }
+  keyring: Keyring
+  portis: any
+  portisAppId: string
+  currentDeviceId: string // wallet id to remove from the keyring when the active wallet changes
```
keepkeyjon:
```suggestion

  /// wallet id to remove from the keyring when the active wallet changes
  currentDeviceId: string
```

so our docs will render nicely, when we get that working again.

packages/hdwallet-portis/src/adapter.ts
```diff
@@ -9,34 +10,43 @@ import Portis from "@portis/web3";
...
+    return new PortisAdapter(keyring, args)
+  }
+
+  public async initialize (): Promise<number> {
+    return Object.keys(this.keyring.wallets).length
+  }
+
+  public async pairDevice (): Promise<HDWallet> {
+    const wallet = this.pairPortisDevice()
+    this.portis.onActiveWalletChanged( async walletAddress => {
```
keepkeyjon: I think the callback belongs in `pairPortisDevice()`, and since the portis object is a singleton, you should query whether you already have it, and then return that one, if it exists.

packages/hdwallet-portis/src/adapter.ts
```diff
@@ -9,34 +10,43 @@ import Portis from "@portis/web3";
...
-      console.log('portis adapter pairDevices')
-      this.portis = this.portis ? this.portis : new Portis(this.portisAppId, 'mainnet')
-      let wallet = new PortisHDWallet(this.portis)
-      await wallet.initialize()
-      const deviceId = await wallet.getDeviceID()
-      this.keyring.add(wallet, deviceId)
-      return wallet
-    }
+  keyring: Keyring
+  portis: any
```
solipsis: As a typescript noob, why any type here?

elmutt: too lazy to figure out portis is supposed to look like

keepkeyjon: in some projects you can elide it, if noImplicitAny is not turned on. not sure about this one.

PR #47, opened 2019-09-30
Refactor the isSomeWallet functions

KeepKeyBrett:
remove _isWallet definitions from wallet interface
allow isWallet to accept any type for its arg

### DIFF
--- a/packages/hdwallet-ledger/src/ledger.ts
+++ b/packages/hdwallet-ledger/src/ledger.ts
@@ -13,7 +13,7 @@ import {
   translateScriptType
 } from './utils'
 
-export function isLedger (wallet: core.HDWallet): wallet is LedgerHDWallet {
+export function isLedger (wallet: any): wallet is LedgerHDWallet {
   return typeof wallet === 'object' && wallet._isLedger === true
 }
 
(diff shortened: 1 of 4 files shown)


packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -13,7 +13,7 @@ import {
   translateScriptType
 } from './utils'
 
-export function isLedger (wallet: core.HDWallet): wallet is LedgerHDWallet {
+export function isLedger (wallet: any): wallet is LedgerHDWallet {
```
keepkeyjon: should go the other way on this, IMO.... makes for better type safety in typescript clients.

packages/hdwallet-trezor/src/trezor.ts
```diff
@@ -47,7 +47,7 @@ import * as Btc from './bitcoin'
 import * as Eth from './ethereum'
 import { TrezorTransport } from './transport'
 
-export function isTrezor(wallet: HDWallet): wallet is TrezorHDWallet {
+export function isTrezor(wallet: any): wallet is TrezorHDWallet {
```
keepkeyjon: likewise

keepkeyjon: and the `isKeepKey` one should be made to take an `HDWallet` instead of an `any`

PR #48, opened 2019-09-30
integration: add test for getPublicKeys

keepkeyjon:


integration/src/bitcoin/bitcoin.ts
```diff
@@ -61,6 +61,59 @@ export function bitcoinTests (get: () => { wallet: HDWallet, info: HDWalletInfo
       expect(await info.btcSupportsCoin('Testnet')).toBeTruthy()
     }, TIMEOUT)
 
+    test.only('getPublicKeys', async () => {
```
keepkeyjon: FIXME: remove the `.only`

integration/src/bitcoin/bitcoin.ts
```diff
@@ -61,6 +61,59 @@ export function bitcoinTests (get: () => { wallet: HDWallet, info: HDWalletInfo
...
+        curve: 'secp256k1',
+        scriptType: BTCInputScriptType.SpendWitness
+      }, {
+        coin: 'Bitcoin',
+        addressNList: bip32ToAddressNList(`m/0'/0'/0'`),
+        curve: 'secp256k1',
+        scriptType: BTCInputScriptType.SpendAddress
+      }, {
+        coin: 'Litecoin',
+        addressNList: bip32ToAddressNList(`m/0'/0'/0'`),
```
keepkeyjon: FIXME: would also be nice to test paths like `m/`, `m/42'`, etc, but the current implementation in `hdwallet-ledger` assumes addressNList has length == 3.

millsmcilroy: Nice to have...

keepkeyjon: we fixed that bug though

PR #49, opened 2019-10-01
Fix ledger.getPublicKeys()

millsmcilroy:
- uses slip44ByCoin() to look up coinDetails
- decouples path and script type by using magic numbers in xpub creation
- fixes operation precedence bug in fingerprint calculation


packages/hdwallet-ledger/src/utils.ts
```diff
@@ -151,7 +156,12 @@ export const networksUtil = {
       bech32: "bc",
       bip32: {
         private: 70615956,
-        public: 70617039
+        public: {
+          p2pkh: 76067358,
```
keepkeyjon: need to make all the coins in this table have the same format.

millsmcilroy: :+1:

keepkeyjon:
Mind adding the integration test for it?

examples/sandbox/index.ts
```diff
@@ -10,6 +10,7 @@ import {
 } from '@shapeshiftoss/hdwallet-core'
 
 import { isKeepKey } from '@shapeshiftoss/hdwallet-keepkey'
+import { isPortis } from '@shapeshiftoss/hdwallet-portis'
```
dviramontes: umm how come this is here ?

millsmcilroy: The portis devs made a breaking change to line 352 (`coin: "Bitcoin"`) when the coin is really Ethereum because they need to pretend it is Bitcoin for the way their wallet works. So I had to add `coin: isPortis(wallet) ? "Bitcoin" : "Ethereum"`

dviramontes: gotcha 👌 / thank you!

keepkeyjon: The real bug here is that we're asking for an xpub with `Ethereum` script type. There is no such thing as an Ethereum xpub. Only a Bitcoin-network-magic xpub with Ethereum's bip44 path.

PR #53, opened 2019-10-02
fix getPublicKeys so 2 calls cant happen simultaenously

elmutt:
fix getPublicKeys so 2 calls cant happen simultaenously


packages/hdwallet-portis/src/portis.ts
```diff
@@ -237,11 +240,17 @@ export class PortisHDWallet implements HDWallet, ETHWallet {
 
   // TODO this needs to handle more than just eth
   public async getPublicKeys(msg: GetPublicKey[]): Promise<PublicKey[]> {
-    const portisResult = await this.portis.getExtendedPublicKey("m/44'/60'/0'")
-    const { result, error } = portisResult
-    if (result) {
-      return [{ xpub: result }]
-    }
+    await this.xpubCallInProgress
```
solipsis: after some reading, I now believe sean is correct and this will not be pre-empted in context of the v8 runtime which i assume is the most common

PR #54, opened 2019-10-03
Memoize the Features object for faster responses

keepkeyjon:


keepkeyjon:
> Spent 5 minutes trying to come up with a better arg name than 'cached' for getFeatures and came up with nothing. LGTM

some prior art, too: https://github.com/shapeshift/hdwallet/blob/70c38f73ff9adfe53ddd6fa6b6a472da3df6a158/packages/hdwallet-trezor/src/trezor.ts#L351

PR #60, opened 2019-10-08
show address on portis if requested

elmutt:
Screen Shot 2019-10-08 at 2 27 05 PM


packages/hdwallet-portis/src/portis.ts
```diff
@@ -301,6 +301,7 @@ export class PortisHDWallet implements HDWallet, ETHWallet {
   }
 
   public async ethGetAddress (msg: ETHGetAddress): Promise<string> {
+    msg.showDisplay === true && this.portis.showPortis()
```
keepkeyjon: please use an `if`

PR #65, opened 2019-10-11
Ledger connection improvements

kaladinlight:
- Handle connect/disconnect events per wallet adapter
- Add open/close functions to ledger transport to manage connection when performing an action against the wallet. This helps handle the fact that the ledger disconnects and reconnects on every entry or exit to an app. The automatic reconnect does not establish a new active connection as the connect needs to be associated with a user event such as click. https://github.com/LedgerHQ/ledgerjs/tree/master/packages/hw-transport-webusb#faq-dom-exception-is-triggered-when-creating-the-transport


packages/hdwallet-keepkey-webusb/src/adapter.ts
```diff
@@ -32,6 +32,8 @@ export class WebUSBKeepKeyAdapter {
   }
 
   private async handleConnectWebUSBKeepKey (e: USBConnectionEvent): Promise<void> {
+    if (e.device.vendorId !== VENDOR_ID) { return }
```
keepkeyjon: mind checking productId too? for webusb KeepKeys, it should be `0x0002`. likewise in the disconnect handler.

packages/hdwallet-ledger/src/bitcoin.ts
```diff
@@ -115,6 +119,8 @@ export async function btcGetAddress (transport: LedgerTransport, msg: BTCGetAddr
 
  */
 export async function btcSignTx (wallet: BTCWallet, transport: LedgerTransport, msg: BTCSignTx): Promise<BTCSignedTx> {
+  await transport.open()
```
keepkeyjon: would it be better to sink the `open()`/`close()` into `transport.call()` so you don't have to remember to add them everywhere?

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -327,7 +327,20 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
...
+    try {
+        var { version } = await this.getFeatures()
+    } catch(err) {
+        if (err.message.includes('0x6d00'))
+            throw new core.WrongApp('Ledger', 'Dashboard')
+
+        console.error(err)
+    }
+
+    await this.transport.close()
```
keepkeyjon: do we need to `close()` from a `finally` clause?

packages/hdwallet-ledger/src/bitcoin.ts
kaladinlight: I tried to do this originally, but there seems to be issues with the user click event bubbling up to that point. I am going to get it working at the HDWallet level for now since it doesn't throw a dom exception at that point and then see what can be done to hoist that further up.

keepkeyjon: ok, SGTM

packages/hdwallet-ledger/src/ledger.ts
kaladinlight: Good catch, thanks!

packages/hdwallet-keepkey-webusb/src/adapter.ts
```diff
@@ -32,6 +32,8 @@ export class WebUSBKeepKeyAdapter {
   }
 
   private async handleConnectWebUSBKeepKey (e: USBConnectionEvent): Promise<void> {
+    if (e.device.vendorId !== VENDOR_ID && e.device.productId !== WEBUSB_PRODUCT_ID) { return }
```
dviramontes:
```suggestion
    if (e.device.vendorId !== VENDOR_ID && e.device.productId !== WEBUSB_PRODUCT_ID) return
```

keepkeyjon: you want `||`, not `&&`, too

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -327,7 +327,20 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
   }
 
   public async getFirmwareVersion (): Promise<string> {
-    const { version } = await this.getFeatures()
+    await this.transport.open()
+
+    var version
```
dviramontes: slightly better (subjective) but it also eliviates the use of _var_ which is not great in JS because of all the scoping rules.  As much as possible we should use _let_ and _const_.
```js
    let version
    try {
        version = (await this.getFeatures()).version
    } catch(err) {
        if (err.message.includes('0x6d00'))
            throw new core.WrongApp('Ledger', 'Dashboard')

        console.error(err)
    } finally {
      await this.transport.close()
    }
```

keepkeyjon: this whole try/catch should go inside `getFeatures()`, no?

kaladinlight: @keepkeyjon yeah that would make more sense, also cleans up the `.version` piece. Fixed.

packages/hdwallet-ledger-webusb/src/transport.ts
```diff
@@ -30,6 +76,15 @@ export class LedgerWebUsbTransport extends LedgerTransport {
     return await getDeviceInfo(this.transport)
   }
 
+  public async open() {
+    const ledgerTransport = await getTransport()
+    this.transport = ledgerTransport
```
dviramontes:
```suggestion
    this.transport = await getTransport()
```

packages/hdwallet-ledger/src/bitcoin.ts
```diff
@@ -146,9 +146,7 @@ export async function btcSignTx (wallet: BTCWallet, transport: LedgerTransport,
     let path = addressNListToBIP32(msg.inputs[i].addressNList)
     let vout = msg.inputs[i].vout
 
-    let tx = await transport.call('Btc', 'splitTransaction', msg.inputs[i].hex,
-      networksUtil[slip44].isSegwitSupported,
-      networksUtil[slip44].areTransactionTimestamped)
+    let tx = await transport.call('Btc', 'splitTransaction', msg.inputs[i].hex, networksUtil[slip44].isSegwitSupported, networksUtil[slip44].areTransactionTimestamped)
```
dviramontes: split by lines easier to read ? (subjective)

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -318,12 +318,22 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
...
-    const { device: { deviceID }} = this.transport as any
+    const { device: { serialNumber: deviceID }} = this.transport as any
     return deviceID
   }
 
   public async getFeatures (): Promise<any> {
-    return this.transport.getDeviceInfo()
+    await this.transport.open()
+    try {
+      return await this.transport.getDeviceInfo()
```
dviramontes: wouldn't the early return prevent us from closing the transport in the final block ?

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -504,15 +533,30 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
...
+    } finally {
+      await this.transport.close()
+    }
   }
 
   public async ethSignMessage (msg: core.ETHSignMessage): Promise<core.ETHSignedMessage> {
-    return eth.ethSignMessage(this.transport, msg)
+    await this.transport.open()
+    try {
+      return await eth.ethSignMessage(this.transport, msg)
```
dviramontes: same questions here about early return

packages/hdwallet-ledger/src/ledger.ts
keepkeyjon: I thought `finally`s ran after `return`s?

keepkeyjon:
```
function tryit() {
  try {
    return console.log("try")
  } finally {
    console.log("finally")
  }
}
tryit()
```
prints:
```
try
finally
```

dviramontes: looks like you are right. I was trying it here too:
https://repl.it/@dviramontes/DistantRecklessKernelmode

dviramontes: disregard @kaladinlight :)

packages/hdwallet-ledger-webusb/src/transport.ts
kaladinlight: haha, yeah I kept thinking I needed to return the promise first. Thanks!

packages/hdwallet-ledger/src/bitcoin.ts
kaladinlight: Going to leave them for now. We can get a linter at some point to make those decisions, but I don't want to revert the change atm.

PR #66, opened 2019-10-11
remove keyring.initializedCount

elmutt:
remove keyring.initializedCount


majorhayes:
You also need to bump the package version.  If there's already a CHANGELOG.md we should update that also.

keepkeyjon:
> You also need to bump the package version
no. don't bump package version from PR's.

PR #67, opened 2019-10-11
fix infinite recursion in keyring.get() call

elmutt:
fix infinite recursion in keyring.get() call


changelog.md
```diff
@@ -1,10 +1,24 @@
...
 and this project adheres to [Semantic Versioning](http://semver.org/).
- 
+
+## [0.11.0] - 2019-10-11
+
+### Added
+
+
+
+### Changed
```
keepkeyjon: I think you can drop the Added/Changed, since we'll be pushing another version here pretty soon ;)

PR #71, opened 2019-10-14
Retry functionality for ledger webusb

ryankeepkey:
- Add `retry` functionality for device navigation errors
- This gives the ability to catch navigation errors when user is on the wrong app by sending error event.
- Adds the ability to "cancel" a request when it is in the retry loop.


packages/hdwallet-ledger/src/utils.ts
```diff
@@ -43,6 +48,38 @@ export function handleError (transport: LedgerTransport, result: any, message: s
   }
 }
 
+export function handleErrorForEvent(result: any, message?: string): any {
+  console.log({ result })
```
dviramontes: do we want to remove this `console.log`

examples/sandbox/index.html
```diff
@@ -89,7 +89,7 @@ <h4>Select</h4>
     <h4>DebugLink</h4>
     <button class='button-outline' id='yes'>Yes</button>
     <button class='button-outline' id='no'>No</button>
-    <button class='button-outline' id='cancel'>Cancel</button>
+    <button class='button-outline cancel'>Cancel</button>
```
keepkeyjon: hmmmm?

packages/hdwallet-ledger-webusb/src/transport.ts
```diff
@@ -86,18 +95,76 @@ export class LedgerWebUsbTransport extends LedgerTransport {
...
+    try {
+      await retry(async (bail) => {
+        if (this.isCancelled) {
+          this.isCancelled = false
+          bail()
+          return
+        }
+
+        // open transport if it's closed
+        await this.open()
```
keepkeyjon: Is `open()` idempotent? what happens for `open(); open();`?

packages/hdwallet-ledger-webusb/src/transport.ts
```diff
@@ -86,18 +95,76 @@ export class LedgerWebUsbTransport extends LedgerTransport {
...
+
+        this.emitEvent({
+          coin: 'none',
+          method: 'retry',
+          response,
+          fromWallet: true,
+          eventType: 'success'
+        })
+      }, {
+        retries: 20,
```
keepkeyjon: what does this do when it hits `20`th retry?

packages/hdwallet-ledger-webusb/src/transport.ts
ryankeepkey: @keepkeyjon Yes as far as I can tell. When open is called multiple times, it opens the current transport and does not cause side effects if it is already open.

packages/hdwallet-ledger-webusb/src/transport.ts
ryankeepkey: After the 20th retry, it will throw the last error encountered. This is just an arbitrary number and can be set to whatever makes sense. It can also be a value that is passed. Not really sure what makes sense until I can test it via axiom

keepkeyjon: cool. let's pull that up in to a SCREAMING_CONSTANT up near the top of the file.

ryankeepkey: lol sounds good

examples/sandbox/index.html
ryankeepkey: @keepkeyjon I just changed this for ease of use. Definitely not married to it at all

packages/hdwallet-ledger/src/utils.ts
```diff
@@ -6,12 +6,13 @@ import {
...
   SelectApp,
   ActionCancelled,
+  NavigateToDashboard,
   makeEvent,
 } from '@shapeshiftoss/hdwallet-core'
 import { LedgerTransport } from './transport'
 import { Buffer } from "buffer";
 
-export function handleError (transport: LedgerTransport, result: any, message: string): void {
+export function handleError (transport: LedgerTransport, result: any, message?: string): void {
```
kaladinlight: Can we have the transport be optional and only emit the error if it exists to avoid duplicating the error handling function?

ryankeepkey: @kaladinlight yes that's a good call. I meant to consolidate this. I will add some logic to throw the error when used outside of an event emitter.

examples/sandbox/index.html
keepkeyjon: was confused by it at first, but now it makes sense... carry on.

examples/sandbox/index.ts
```diff
@@ -188,6 +188,16 @@ async function deviceConnected (deviceId) {
   keyring.on(['*', '*', Events.DISCONNECT], async (deviceId) => {
     $keyring.find(`option[value="${deviceId}"]`).remove()
   })
+
+  keyring.on(['Ledger', '0001', 'ledger.none.retry.response'], async ([ deviceId, event ]) => {
```
keepkeyjon: side thought: I think we ought to make this a "first class" event, like `Events.CONNECT`. don't need to do it as part of this PR though.

examples/sandbox/index.ts
```diff
@@ -188,6 +188,16 @@ async function deviceConnected (deviceId) {
...
+    $ledgerError.append('<span>' + event.message.response + '</span>')
+    document.getElementById('#ledgerModal').className = 'modale opened'
+  })
+
+  keyring.on(['Ledger', '0001', 'ledger.none.retry.success'], async ([ deviceId, event ]) => {
```
keepkeyjon: likewise

dviramontes:
superseded by #86

PR #72, opened 2019-10-15
tweak: Add now

dviramontes:
let's try it

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,2 @@
+yarn --ignore-optional
+yarn build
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -0,0 +1,2 @@
+yarn --ignore-optional
```
dviramontes: may or may not need a shebang here

build.sh
```diff
@@ -0,0 +1,2 @@
...
+yarn build
```
keepkeyjon:
```suggestion
yarn now-build
```

PR #81, opened 2019-10-17
change var keepkey > wallet

KeepKeyBrett:
rename keepkey var to wallet to avoid confusion. this var could be any wallet in the keyring


packages/hdwallet-core/src/keyring.ts
```diff
@@ -54,8 +54,8 @@ export class Keyring extends eventemitter2.EventEmitter2 {
     if (!this.get(deviceID)) return
 
     try {
-      const keepkey = this.get(deviceID)
-      await keepkey.disconnect()
+      const wallet = this.get(deviceID)
```
keepkeyjon: can you add a `if (!wallet) return` in between those?

PR #83, opened 2019-10-21
fix: try .create() if .openConnected() throws an error in ledger transport

millsmcilroy:
Also, it is better to call create() instead of request() because request will always pop up the webusb permission modal, even if the user has already given that device permission.

### DIFF
--- a/packages/hdwallet-ledger-webusb/src/transport.ts
+++ b/packages/hdwallet-ledger-webusb/src/transport.ts
@@ -48,13 +48,22 @@ export async function getTransport(): Promise<TransportWebUSB> {
 
   let ledgerTransport
   try {
-    ledgerTransport = await TransportWebUSB.openConnected() || await TransportWebUSB.request()
+    ledgerTransport = await TransportWebUSB.openConnected()
   } catch (err) {
     if (err.name === 'TransportInterfaceNotAvailable') {
       throw new ConflictingApp('Ledger')
     }
+  }
 
-    throw new WebUSBCouldNotPair('Ledger', err.message)
+  if (!ledgerTransport) {
+    try {
+      ledgerTransport = await TransportWebUSB.create()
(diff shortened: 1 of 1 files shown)


packages/hdwallet-ledger-webusb/src/transport.ts
```diff
@@ -48,13 +48,22 @@ export async function getTransport(): Promise<TransportWebUSB> {
 
   let ledgerTransport
```
keepkeyjon: Suggestion:

```
  try {
    let transport = await TransportWebUSB.openConnected()
    if (transport) {
      return transport
    }

    // Use create() rather than request() to avoid the WebUSB pairing dialog when possible.
    return await TransportWebUSB.create()
  } catch (err) {
    if (err.name === 'TransportInterfaceNotAvailable') {
      throw new ConflictingApp('Ledger')
    }
    throw new WebUSBCouldNotPair('Ledger', err.message)
  }
```

kaladinlight:
Oh cool. It looks like create actually does what `openConnected()` was trying to accomplish too? I just tested out using only create and it will show the pair popup on initial permission, and then subsequent calls will just pair to the first ledger device it finds that has been given permission at that point. Want to test that out too on your end to verify and then we can strip the first try block?

packages/hdwallet-ledger-webusb/src/transport.ts
kaladinlight: @keepkeyjon Part of the initial problems was that if `openConnected` threw an error, it would short circuit the logic to the `catch` block, resulting in `create/request` never being called as a fallback. See my comment below though in regards to `create` functionality which should make this all much cleaner anyways.

millsmcilroy:
Closing this PR and moving these changes to https://github.com/shapeshift/hdwallet/pull/86

PR #87, opened 2019-10-24
Handle app navigation

kaladinlight:
Use timeout to determine if usb events are due to app navigation or actual connecting/disconnecting of the device.

This works because for the app navigation events, we are always presented with a disconnect event immediately followed by a connect event. Therefore, if we add the sleep in `handleDisconnect` and give enough time for the connect event to be received and timestamp tracked, we can be sure it was due to app navigation. (we really need to stress test this hard for any edge cases. nobody likes magic sleeps in code, but I can't think of a better way to accomplish this as of right now and even though it's ugly, it seems to provide us the desired behavior)

Connect is run regardless to re-establish the connection to the webusb transport. It would appear that because this is happening on a usb window event, it seems to be being treated as a "user event" and allows for the automatic transport reconnect? Only way I can explain it right now. 

It would appear the open/close around the call might be a fail just due to call stack size on various calls and if the "user event" is still visible at the time `open` is called (which it seems like there are various cases where this isn't true). Note, the open/close behavior attempt was tested on the vanilla branch without any of Ryan's retry logic in place. There did seem to be a difference when the retry logic was added and how open/close was being treated.

### DIFF
--- a/packages/hdwallet-ledger-webusb/src/adapter.ts
+++ b/packages/hdwallet-ledger-webusb/src/adapter.ts
@@ -6,7 +6,7 @@ import TransportWebUSB from '@ledgerhq/hw-transport-webusb'
 const VENDOR_ID = 11415
 const APP_NAVIGATION_DELAY = 1000
 
-function timeout(ms): Promise<void> {
+function timeout(ms: number): Promise<void> {
     return new Promise(resolve => setTimeout(resolve, ms));
 }
 
--- a/packages/hdwallet-ledger-webusb/src/adapter.ts
+++ b/packages/hdwallet-ledger-webusb/src/adapter.ts
@@ -5,8 +5,18 @@ import TransportWebUSB from '@ledgerhq/hw-transport-webusb'
 
 const VENDOR_ID = 11415
 
+function timeout(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+async function sleep(fn, ...args) {
+    await timeout(3000);
+    return fn(...args);
+}
+
 export class WebUSBLedgerAdapter {
   keyring: Keyring
+  connectTimestamp: number = 0
 
   constructor(keyring: Keyring) {
     this.keyring = keyring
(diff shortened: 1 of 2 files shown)


packages/hdwallet-ledger-webusb/src/adapter.ts
```diff
@@ -5,8 +5,18 @@ import TransportWebUSB from '@ledgerhq/hw-transport-webusb'
 
 const VENDOR_ID = 11415
 
+function timeout(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+async function sleep(fn, ...args) {
```
keepkeyjon: unused?

packages/hdwallet-ledger-webusb/src/adapter.ts
```diff
@@ -5,8 +5,18 @@ import TransportWebUSB from '@ledgerhq/hw-transport-webusb'
...
+    await timeout(3000);
```
ryankeepkey: ?

packages/hdwallet-ledger-webusb/src/adapter.ts
kaladinlight: yes removed.thx

packages/hdwallet-ledger-webusb/src/adapter.ts
kaladinlight: removed

packages/hdwallet-ledger-webusb/src/adapter.ts
```diff
@@ -5,8 +5,13 @@ import TransportWebUSB from '@ledgerhq/hw-transport-webusb'
 
 const VENDOR_ID = 11415
 
+function timeout(ms) {
```
keepkeyjon:
```suggestion
function timeout(ms) : Promise<void> {
```

packages/hdwallet-ledger-webusb/src/adapter.ts
```diff
@@ -24,16 +29,27 @@ export class WebUSBLedgerAdapter {
   private async handleConnectWebUSBLedger(e: USBConnectionEvent): Promise<void> {
     if (e.device.vendorId !== VENDOR_ID) return
 
+    this.connectTimestamp = e.timeStamp
+
+    await timeout(1000) // timeout gives time to detect if it is an app navigation based disconnec/connect event
```
keepkeyjon: Should the number itself go in some `const SCREAMING_GLOBAL = 1000` ?

packages/hdwallet-ledger-webusb/src/adapter.ts
kaladinlight: Oh right, typescript :)

packages/hdwallet-ledger-webusb/src/adapter.ts
kaladinlight: Sure thing lol

PR #90, opened 2019-10-28
Portis btc

elmutt:


packages/hdwallet-portis/src/portis.ts
```diff
@@ -82,11 +203,90 @@ export function isPortis(wallet: HDWallet): wallet is PortisHDWallet {
...
+  // btc stuff
+
+
+  public async btcGetAddress (msg: BTCGetAddress): Promise<string> {
+    
+    const change = msg.addressNList[3]
+    const index = msg.addressNList[4]
+
+    const b32string = addressNListToBIP32(msg.addressNList)
+    const hardPath = b32string.slice(0, b32string.lastIndexOf(`'`)+1)
```
keepkeyjon: Don't do string math. Use `hardenedPath` to get the hardened part of the addressNList.

Also, don't assume that the relpath will always include 2 entries.

packages/hdwallet-portis/src/portis.ts
```diff
@@ -82,11 +203,90 @@ export function isPortis(wallet: HDWallet): wallet is PortisHDWallet {
...
 
-export class PortisHDWallet implements HDWallet, ETHWallet {
+export class PortisHDWallet implements HDWallet, ETHWallet, BTCWallet {
+
+  // btc stuff
+
+
+  public async btcGetAddress (msg: BTCGetAddress): Promise<string> {
+    
+    const change = msg.addressNList[3]
```
keepkeyjon: Don't assume any structure on the `addressNList` passed in.

keepkeyjon: also don't assume `addressNList.length > 0`. If you need to assume so, throw an error in those cases.... but I'd prefer to make this as generic as possible, agnostic to paths.

packages/hdwallet-portis/src/bitcoin.ts
```diff
@@ -0,0 +1,280 @@
...
+export async function btcGetAddress (msg: BTCGetAddress, portis: any): Promise<string> {
+
+  if(!msg.addressNList.length)
+    throw new Error('Empty addressNList')
+
+  const scriptType = msg.scriptType
+  const purpose = msg.addressNList[0]
+
+  const hardPath = hardenedPath(msg.addressNList)
+  const hardPathString = 'm/'+hardPath.reduce( (acc, val, i) => (i !== hardPath.length - 1) ? acc + (val - 2147483648) + `'/` : acc + (val - 2147483648) + `'`, '')
```
keepkeyjon:
```suggestion
  const hardPathString = addressNListToBip32(hardPath)
```

packages/hdwallet-portis/src/bitcoin.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  const { result: xpub } = await portis.getExtendedPublicKey(hardPathString, "Bitcoin")
+
+  const relPath = relativePath(msg.addressNList)
+  const relPathString = relPath.reduce( (acc, val, i) => (i !== relPath.length - 1) ? acc + val + '/' : acc + val, '')
```
keepkeyjon:
```suggestion
  const relPathString = addressNListToBip32(relPath)
```

keepkeyjon: and actually, you don't even need it as a string anyway... just pass relPath where you're passing relPathString

packages/hdwallet-portis/src/bitcoin.ts
```diff
@@ -0,0 +1,280 @@
...
+  return {
+    coin,
+    scriptType: BTCInputScriptType.SpendWitness,
+    addressNList: [ 0x80000000 + 84, 0x80000000 + slip44, 0x80000000 + accountIdx ]
+  }
+}
+
+
+export function btcNextAccountPath (msg: BTCAccountPath): BTCAccountPath | undefined {
+  let description = describeUTXOPath(msg.addressNList, msg.coin, msg.scriptType)
```
keepkeyjon: Does portis support more than one BTC account? If not, this whole function should be `return undefined`

packages/hdwallet-portis/src/bitcoin.ts
```diff
@@ -0,0 +1,280 @@
...
+    paths = paths.filter(path => { return path.scriptType === msg.scriptType })
+
+  return paths
+}
+
+export function btcSupportsScriptType (coin: Coin, scriptType: BTCInputScriptType): Promise<boolean> {
+
+  console.log('btcSupportsScriptType!!', {coin, scriptType})
+  if(coin !== 'Bitcoin')
+    return Promise.resolve(false)
```
keepkeyjon: if you make the function `async`, you can just `return false` here. a little cleaner.

packages/hdwallet-portis/src/portis.ts
```diff
@@ -251,8 +190,10 @@ export class PortisHDWallet implements HDWallet, ETHWallet {
           console.error(e)
       }
       for (let i = 0; i < msg.length; i++) {
-        const { addressNList } = msg[i];
-        const portisResult = await this.portis.getExtendedPublicKey(addressNListToBIP32(addressNList))
+        const { addressNList, coin } = msg[i];
+        const slip44 = 0x80000000 + slip44ByCoin(coin)
+        // TODO we really shouldnt be every using the "bitcoin" string parameter but is here for now to make it work with their btc address on their portis wallet.
+        const portisResult = await this.portis.getExtendedPublicKey(addressNListToBIP32(addressNList), addressNList[1] === slip44 ? 'Bitcoin' : '')
```
keepkeyjon: What I suggested here may have broken it for ETH, since `slip44ByCoin('Ethereum')` will give you `60'`, and that will match when given a standard ETH path, causing the ternay to say `Bitcoin`.... which IIUC, isn't what you want.

keepkeyjon: I think what you want instead is `const slip44 = slip44ByCoin('Bitcoin')`

packages/hdwallet-portis/src/bitcoin.ts
elmutt: doing addressNListToBip32(relPath) here returns the string `m/0/0` instead of `0/0`

passing `m/0/0` to derivePath causes the error:

```
ncaught (in promise) TypeError: Expected master, got child
    at BIP32.derivePath (bip32.js:168)
    at Object.<anonymous> (bitcoin.ts:153)
    at step (index.js:26)
    at Object.next (index.js:26)
    at fulfilled (index.js:26)
```

packages/hdwallet-portis/src/bitcoin.ts
elmutt: ok

packages/hdwallet-portis/src/bitcoin.ts
elmutt: k

packages/hdwallet-portis/src/portis.ts
elmutt: how about this:

```
        const bitcoinSlip44 = 0x80000000 + slip44ByCoin('Bitcoin')
        const portisResult = await this.portis.getExtendedPublicKey(addressNListToBIP32(addressNList), addressNList[1] === bitcoinSlip44 ? 'Bitcoin' : '')
```

elmutt: ya.  that

packages/hdwallet-portis/src/bitcoin.ts
elmutt: Also cant pass relPath instead of relPathString into derivePath() because it expects a string not an array

packages/hdwallet-portis/src/portis.ts
keepkeyjon: perfect

packages/hdwallet-portis/src/bitcoin.ts
keepkeyjon: hm. thought it supported that

keepkeyjon: maybe `addressNListToBip32(relPath).substr(2, -1)`?

elmutt: ya thats probably easier to read

packages/hdwallet-portis/src/bitcoin.ts
elmutt: good call.  i thought i had this returning undefined already...

packages/hdwallet-portis/src/bitcoin.ts
elmutt: I think we want:

 const relPathString = addressNListToBIP32(relPath).substr(2)

PR #92, opened 2019-10-30
fix: nano x get firmware version call

dviramontes:
as titled
- looks like the call to ledger-live common returns diff payloads based on the device type
`this.transport.getDeviceInfo()`

- also updates ledger deps to latest release since [i noticed the version call fix on this release](https://github.com/LedgerHQ/ledger-live-common/releases/tag/v8.0.0) (happy to make that a separate PR if we are concerned about other breaking changes..

### DIFF
--- a/packages/hdwallet-ledger/src/ledger.ts
+++ b/packages/hdwallet-ledger/src/ledger.ts
@@ -323,12 +323,9 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
   }
 
   public async getFeatures (): Promise<any> {
-    const res = await this.transport.getDeviceInfo()
+    const res = await this.transport.call(null, 'getDeviceInfo')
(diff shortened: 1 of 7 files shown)


packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -325,7 +325,10 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
   public async getFeatures (): Promise<any> {
     const res = await this.transport.getDeviceInfo()
     handleError(res, this.transport)
```
keepkeyjon: The `res.payload` stuff, and the `handleError()` call are both assuming that the `getDeviceInfo()` call went through `this.transport.call()`. It would be better if that end was fixed, so then we get these messages and the responses when logging from `keyring.onAny(...)`, rather than changing this end of it.

PR #95, opened 2019-10-31
fix: better guards for is object

dviramontes:
[use lodash.isObject for object checks](https://lodash.com/docs/4.17.15#isObject)
to prevent stuff like this
```
typeof null === 'object'
true
```
from happening

testing: 
`./node_modules/.bin/jest -t  'wallet : guards'`

### DIFF
--- a/packages/hdwallet-core/src/wallet.ts
+++ b/packages/hdwallet-core/src/wallet.ts
@@ -10,6 +10,7 @@ import {
 } from './ethereum'
 import { DebugLinkWallet } from './debuglink'
 import { Transport } from './transport';
+import { isObject } from 'lodash';
 
 export type BIP32Path = Array<number>
 
@@ -113,11 +114,11 @@ export type Coin = string
  ```
  */
 export function supportsBTC(wallet: any): wallet is BTCWallet {
-  return typeof wallet === 'object' && wallet._supportsBTC === true
+  return isObject(wallet) && wallet._supportsBTC === true
 }
 
 export function infoBTC(info: any): info is BTCWalletInfo {
-  return typeof info === 'object' && info._supportsBTCInfo === true
+  return isObject(info) && info._supportsBTCInfo === true
 }
 
 /**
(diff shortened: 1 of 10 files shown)


packages/hdwallet-core/src/wallet.ts
```diff
@@ -113,11 +114,11 @@ export type Coin = string
  ```
  */
 export function supportsBTC(wallet: any): wallet is BTCWallet {
-  return typeof wallet === 'object' && wallet._supportsBTC === true
+  return isObject(wallet) && (wallet as HDWallet)._supportsBTC === true
```
kaladinlight: Is there a reason it worked without the type case before? Or is this just an extra safe guard?

packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -62,7 +62,7 @@ import * as Eth from "./ethereum"
 import { KeepKeyTransport } from "./transport";
 
 export function isKeepKey(wallet: HDWallet): wallet is KeepKeyHDWallet {
-  return typeof wallet === 'object' && (wallet as any)._isKeepKey === true
+  return isObject(wallet) && (wallet as any)._isKeepKey === true
```
kaladinlight: Would it make sense to type case the rest of the `wallet` to `HDWallet` instead of any as well?

packages/hdwallet-core/src/wallet.ts
majorhayes: because `wallet` is `any`, but `isObject` has a return type of `wallet is object`, so at that point TypeScript sees `wallet` as `object`, not as `any`.  `_supportsBTC` is not a property available on `object`.

kaladinlight: Interesting. Thanks for the explanation!

packages/hdwallet-keepkey/src/keepkey.ts
dviramontes: that would make sense but i would be worried about introducing side-effects to other wallets, maybe separate PR ?

PR #96, opened 2019-11-01
Add Ledger validateCurrentApp  and openApp() functionality

millsmcilroy:


packages/hdwallet-ledger-webusb/src/transport.ts
```diff
@@ -91,6 +102,9 @@ export class LedgerWebUsbTransport extends LedgerTransport {
       }))
       if (coin) {
         response = await new (translateCoin(coin))(this.transport)[method](...args)
+      } else if (method === 'openApp') {
+        const appName = args[0]
```
dviramontes: another way to write this
```
const [ appName ] = args
```

millsmcilroy: I wonder what the TypeScript way is

dviramontes: me too! i looked around and i think its more to do in the assignment
```typescript
const args: number[] = [1, 2, 3]
const [ one, two, three ] = args
```

packages/hdwallet-ledger-webusb/src/transport.ts
```diff
@@ -91,6 +102,9 @@ export class LedgerWebUsbTransport extends LedgerTransport {
       }))
       if (coin) {
         response = await new (translateCoin(coin))(this.transport)[method](...args)
+      } else if (method === 'openApp') {
+        const [ appName ] = args
+        response = await openApp(this.transport, appName)
       } else {
         response = await (translateMethod(method))(this.transport)
```
kaladinlight: Can we just use the else and spread the args if there are any?

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -329,6 +330,38 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
...
+    return res.payload
+  }
+
+  /**
+   * Returns Ledger's name for an app given a coin's
+   * ticker symbol
+   * @param symbol - i.e. 'BCH'
+   * @returns appName i.e. "Bitcoin Cash"
+   */
+  public getAppNameBySymbol (symbol: string): string {
```
keepkeyjon: where is this used?

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -329,6 +330,38 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
...
+    return appsUtil[symbol]
+  }
+
+  /**
+   * Prompt user to open given app on device
+   * User must be in dashboard
+   * @param appName - human-readable app name i.e. "Bitcoin Cash"
+   */
+  public async openApp (appName: string): Promise<void> {
```
kaladinlight: Wondering if we should just pass the symbol to this function and let it determine the app name since the lookup table is in here anyways

kaladinlight: We should also throw an error for symbols we don't have mapped to an app name and toUpperCase the symbol name in case it is passed with different casing

packages/hdwallet-ledger/src/ledger.ts
millsmcilroy: https://beta.shapeshift.com/

packages/hdwallet-ledger/src/ledger.ts
keepkeyjon: IMO these should be the HDWallet name for a coin so the interface across all of HDWallet is uniform. A.k.a names from this list: https://github.com/shapeshift/hdwallet/blob/ba223f94dca0797eef3bb5255cc296a30a0329de/packages/hdwallet-keepkey/src/bitcoin.ts#L44. In `coinMap`, on the shapeshift.com side of things, that is called `keepkeyName`, FWIW. (needs to be renamed `hdwalletName` or something)

keepkeyjon: i.e. this should be `public async openApp(coin: Coin): Promise<void>`

packages/hdwallet-ledger/src/ledger.ts
keepkeyjon: ohh, it's used in `openApp`

packages/hdwallet-ledger-webusb/src/transport.ts
millsmcilroy: Had to add a // @ts-ignore

packages/hdwallet-ledger/src/ledger.ts
millsmcilroy: I add a validateCurrentApp method

PR #105, opened 2019-11-11
Ledger disconnected error

kaladinlight:
Add exception for ledger's `DisconnectedDeviceDuringOperation` error.
Remove open/close which is no longer used.
Keep U2F package updated with WebUSB.


packages/hdwallet-ledger-u2f/src/transport.ts
```diff
@@ -15,7 +17,9 @@ function translateCoin(coin: string): (any) => void {
 
 function translateMethod(method: string): (any) => void {
   return {
+    'getAppAndVersion': getAppAndVersion,
```
dviramontes: 👍

PR #108, opened 2019-11-13
Ledger ETH support for BIP44 account paths

kaladinlight:


packages/hdwallet-portis/src/portis.ts
```diff
@@ -181,7 +181,7 @@ export class PortisHDWallet implements HDWallet, ETHWallet, BTCWallet {
     return this.info.describePath(msg)
   }
 
-  public async getPublicKeys(msg: GetPublicKey[]): Promise<PublicKey[]> {
```
elmutt: Why change portis?

kaladinlight: The function definition in the HDWallet interface changed.

elmutt: less risk to just leave it

kaladinlight: Should behave exactly the same way.

elmutt: It changed adding `| null`

no need to touch portis please

kaladinlight: Right, it matches HDWallet definition in core.

elmutt: its less risk to not touch this line

keepkeyjon: @elmutt it'll be fine, no risk. Types in typescript are just an annotation that the compiler checks, so if it builds, it's good.

packages/hdwallet-ledger/src/ethereum.ts
```diff
@@ -1,45 +1,84 @@
...
   const res = await transport.call('Eth', 'getAddress', bip32path, !!msg.showDisplay)
   handleError(res, transport, 'Unable to obtain ETH address from device.')
 
   return res.payload.address
 }
 
-export async function ethSignTx(transport: LedgerTransport, msg: ETHSignTx): Promise<ETHSignedTx> {
-  const bip32path = addressNListToBIP32(msg.addressNList)
+// Adapted from https://github.com/LedgerHQ/ledger-wallet-webtool
+export async function ethGetPublicKeys (transport: LedgerTransport, msg: Array<core.GetPublicKey>): Promise<Array<core.PublicKey>> {
```
keepkeyjon:
```suggestion
export async function ethGetPublicKeys (transport: LedgerTransport, msg: Array<core.GetPublicKey>): Promise<Array<core.PublicKey | null>> {
```

packages/hdwallet-ledger/src/bitcoin.ts
```diff
@@ -79,6 +64,56 @@ export async function btcGetAddress (transport: LedgerTransport, msg: BTCGetAddr
   return res.payload.bitcoinAddress
 }
 
+// Adapted from https://github.com/LedgerHQ/ledger-wallet-webtool
+export async function btcGetPublicKeys (transport: LedgerTransport, msg: Array<core.GetPublicKey>): Promise<Array<core.PublicKey>> {
```
keepkeyjon:
```suggestion
export async function btcGetPublicKeys (transport: LedgerTransport, msg: Array<core.GetPublicKey>): Promise<Array<core.PublicKey | null>> {
```

packages/hdwallet-ledger/src/ethereum.ts
```diff
@@ -84,22 +123,26 @@ export function ethSupportsNativeShapeShift(): boolean {
   return false
 }
 
-export function ethGetAccountPaths (msg: ETHGetAccountPath): Array<ETHAccountPath> {
+export function ethGetAccountPaths (msg: core.ETHGetAccountPath): Array<core.ETHAccountPath> {
+  /* TODO support both ETH account paths
```
keepkeyjon: You might have to comment out a corresponding test case.

packages/hdwallet-ledger/src/bitcoin.ts
```diff
@@ -79,6 +64,56 @@ export async function btcGetAddress (transport: LedgerTransport, msg: BTCGetAddr
...
+      publicKey = compressPublicKey(publicKey)
+
+      const coinDetails: any = networksUtil[core.slip44ByCoin(coin)]
+      const childNum: number = addressNList[addressNList.length -1]
+
+      scriptType = scriptType || core.BTCInputScriptType.SpendAddress
+      const networkMagic = coinDetails.bitcoinjs.bip32.public[scriptType]
+
+      const xpub = createXpub(
+        3,
```
keepkeyjon:
```suggestion
        addressNList.length,
```

packages/hdwallet-ledger/src/ethereum.ts
```diff
@@ -1,45 +1,84 @@
...
+
+      let { payload: { publicKey, chainCode } } = res2
+      publicKey = compressPublicKey(publicKey)
+
+      const coinDetails: any = networksUtil[core.slip44ByCoin(coin)]
+      const childNum: number = addressNList[addressNList.length -1]
+      const networkMagic = coinDetails.bitcoinjs.bip32.public[scriptType]
+
+      const xpub = createXpub(
+        3,
```
keepkeyjon:
```suggestion
        addressNList.length,
```

keepkeyjon: (this was broken before your change, but that's what it really should be)

packages/hdwallet-core/src/wallet.ts
```diff
@@ -19,7 +19,8 @@ export interface GetPublicKey {
   showDisplay?: boolean,
   scriptType?: BTCInputScriptType,
   curve: string,
-  coin?: Coin
+  coin?: Coin,
+  symbol?: Symbol
```
keepkeyjon: You should key this off of what the `coin` is, instead of adding another one.

kaladinlight: Symbol is already on the payload we are passing in which is why I added that as an optional key

keepkeyjon: yeah, I see why you did that now... lemme think about it for a bit.

packages/hdwallet-ledger/src/ethereum.ts
kaladinlight: I added one to just test the single path and added skip to the other one until the non bip44 is supported.

keepkeyjon: 👍

packages/hdwallet-ledger/src/ethereum.ts
```diff
@@ -1,45 +1,84 @@
...
-export async function ethSignTx(transport: LedgerTransport, msg: ETHSignTx): Promise<ETHSignedTx> {
-  const bip32path = addressNListToBIP32(msg.addressNList)
+// Adapted from https://github.com/LedgerHQ/ledger-wallet-webtool
+export async function ethGetPublicKeys (transport: LedgerTransport, msg: Array<core.GetPublicKey>): Promise<Array<core.PublicKey>> {
+    const xpubs = []
+
+    for (const getPublicKey of msg) {
+      let { addressNList, coin, symbol, scriptType } = getPublicKey
+
+      if (symbol !== 'ETH') {
```
keepkeyjon: I think this test should match what their device actually cares about, which I think is that `addressNList` must start with  `[0x80000000 + 44, 0x80000000 + 60]` if you're using the Eth app. Then we can avoid adding `symbol` to `GetPublicKey`, which seems odd (not sure why it's being passed in that other spot, but it shouldn't be what we're using here).

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -389,54 +380,20 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
...
 
-      const xpub = createXpub(
-        3,
-        fingerprint,
-        childNum,
-        chainCode,
-        publicKey,
-        networkMagic
-      )
+    const { name } = res.payload
```
dviramontes:
```suggestion
    const { payload: { name } } = res
```

packages/hdwallet-ledger/src/ethereum.ts
```diff
@@ -72,7 +111,7 @@ export async function ethSignTx(transport: LedgerTransport, msg: ETHSignTx): Pro
     v: parseInt(v, 16),
     r: '0x' + r,
     s: '0x' + s,
-    serialized: '0x' + toHexString(tx.serialize())
+    serialized: '0x' + core.toHexString(tx.serialize())
```
dviramontes:
```suggestion
    serialized: `0x${core.toHexString(tx.serialize())}`
```

packages/hdwallet-ledger/src/ethereum.ts
kaladinlight: Decided to leave the tests and `ethGetAccountPaths` alone (including both ledger live and chrome derivation paths). I will just add the length to the `ethGetPublicKeys` filter to only include BIP44 account path for now.

keepkeyjon: even better

packages/hdwallet-ledger/src/ethereum.ts
kaladinlight: This syntax is spread throughout the whole code base. Not going to worry about this one spot at this time.

packages/hdwallet-ledger/src/ethereum.ts
```diff
@@ -1,45 +1,85 @@
...
-  const bip32path = addressNListToBIP32(msg.addressNList)
+// Adapted from https://github.com/LedgerHQ/ledger-wallet-webtool
+export async function ethGetPublicKeys (transport: LedgerTransport, msg: Array<core.GetPublicKey>): Promise<Array<core.PublicKey | null>> {
+    const xpubs = []
+
+    for (const getPublicKey of msg) {
+      let { addressNList, coin, scriptType } = getPublicKey
+
+      // Only get public keys for ETH account paths
+      if (!addressNList.includes(0x80000000 + 44, 0) || !addressNList.includes(0x80000000 + 60, 1)) {
```
keepkeyjon: woah, didn't know about that second arg to `includes()`, nice!

keepkeyjon:
@kaladinlight mind adding release notes in changelog.md?

PR #109, opened 2019-11-14
ledger: get app list

dviramontes:


packages/hdwallet-ledger-webusb/src/transport.ts
```diff
@@ -14,6 +14,7 @@ import Btc from '@ledgerhq/hw-app-btc'
 import getAppAndVersion from '@ledgerhq/live-common/lib/hw/getAppAndVersion'
 import getDeviceInfo from '@ledgerhq/live-common/lib/hw/getDeviceInfo'
 import openApp from '@ledgerhq/live-common/lib/hw/openApp'
+import { listApps } from "@ledgerhq/live-common/lib/apps/hw"
```
kaladinlight: What is the difference between this `apps/hw` implementation and the `/hw` implementation here: https://github.com/LedgerHQ/ledger-live-common/blob/master/src/hw/listApps.js

dviramontes: Good questions. @ryankeepkey mentioned this. The one at  https://github.com/LedgerHQ/ledger-live-common/blob/master/src/hw/listApps.js uses the native transport vs the one we are using we pass it our transport. Not sure what the difference is beyond that but something to keep in mind the one we are using here doesn't work.

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -353,6 +354,13 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
     }
   }
 
+  public async listApps (): Promise<any> {
+    const deviceInfo = await getDeviceInfo(this.transport)
+    const res = await this.transport.call(null, 'listApps', deviceInfo)
+    console.log({ res })
```
keepkeyjon:
```suggestion
```

packages/hdwallet-ledger/src/ledger.ts
```diff
@@ -353,6 +354,13 @@ export class LedgerHDWallet implements core.HDWallet, core.BTCWallet, core.ETHWa
     }
   }
 
+  public async listApps (): Promise<any> {
```
keepkeyjon: Is it possible to be more specific with the return type? i.e. `Promise<string[]>` or something?

packages/hdwallet-ledger-webusb/src/transport.ts
kaladinlight: I am pretty sure both of them pass the transport. That is the way we set up those calls. We wouldn't need to do the `getDeviceInfo` with the `/hw` either. The rest of the utility functions we are using are also in `/hw`. I was surprised to even see the other implementation of `listApps` in `apps/hw`.

BitHighlander:
stale

PR #110, opened 2019-11-15
fix ledger app navigation timing issue

ryankeepkey:
- Makes the app navigation connect/reconnect more manageable.


dviramontes:
Verified patch works for Nano X 👍

PR #142, opened 2020-04-22
Eos

markrypto:


packages/hdwallet-keepkey/src/eos.ts
```diff
@@ -0,0 +1,204 @@
...
+  if (c.charCodeAt(0) >= "a".charCodeAt(0) && c.charCodeAt(0) <= "z".charCodeAt(0)) {
+    return ((c.charCodeAt(0) - "a".charCodeAt(0)) + 6)
+  }
+  if (c.charCodeAt(0) >= "1".charCodeAt(0) && c.charCodeAt(0) <= "5".charCodeAt(0)) {
+    return ((c.charCodeAt(0) - "1".charCodeAt(0)) + 1)
+  }
+  return 0
+}
+
+function nameToNumber(name:string) : string {
```
elmutt, line 44: Whats this do?

packages/hdwallet-keepkey/src/eos.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    } else { // last iteration
+        c = c.and(Long.fromString("0f", true, 16))
+    }    
+      value = value.or(c)
+  }
+  return value.toString()
+}
+
+function assetToNumber(asset:string): [number, number] {
```
elmutt, line 66: whats this do?

packages/hdwallet-keepkey/src/eos.ts
markrypto: Converts the alphanumeric name of an action to a format that is appropriate for signing. This was derived from the python-keepkey tests for eos that came from trezor.

packages/hdwallet-keepkey/src/eos.ts
markrypto: Same as the nameToNumber(), converts the asset name to a number for signing. Derived from the trezor reference for testing.

PR #159, opened 2020-06-06
Use pinCached to check if device is locked with PIN

### LINKED ISSUE #158: [KeepKey] `isLocked` returns false even if device is locked with PIN
[`isLocked`](https://github.com/shapeshift/hdwallet/blob/master/packages/hdwallet-keepkey/src/keepkey.ts#L689) always returns `false`, even when the device is locked with a PIN (passphrase not tested). Looks like there is a bug on this line: https://github.com/shapeshift/hdwallet/blob/master/packages/hdwallet-keepkey/src/keepkey.ts#L691

`features.pinProtection && !features.pinProtection` always evaluates to false, since `pinProtection` is never truthy and falsy at the same time. This should probably be changed to `features.pinProtection && !features.pinCached`.

Mrtenz:
Fixes #158


greatwolf:
Is there a reason why this PR hasn't been reviewed and merged in? This is a trivial bugfix. It should have happened a long time ago already.

Mrtenz:
I have no clue. If you want a quick workaround though, you can just write your own `isLocked` function and use the implementation from this PR.

greatwolf:
Thanks, I'm mainly commenting this to get the maintainers attention.
The lack of action from them gives the impression that their products and libraries have been abandoned.

BitHighlander (approved):

PR #161, opened 2020-06-19
Add native

SterL:


packages/hdwallet-core/src/bitcoin.ts
```diff
@@ -212,10 +213,171 @@ export interface BTCWalletInfo {
 }
 
 export interface BTCWallet extends BTCWalletInfo {
-  _supportsBTC: boolean;
```
kaladinlight, line 215: Let's add this back just in case (I removed since there were dupes, but I don't want to accidentally break logic around this)

packages/hdwallet-native/src/bitcoin.ts
```diff
@@ -0,0 +1,293 @@
...
+      return undefined;
+    }
+  };
+}
+
+export function MixinNativeBTCWallet<TBase extends core.Constructor>(
+  Base: TBase
+) {
+  return class MixinNativeBTCWallet extends Base {
+    seed: Buffer;
```
kaladinlight, line 141: And we can add it here as `true`

packages/hdwallet-native/src/native.ts
```diff
@@ -0,0 +1,184 @@
...
+    return false;
+  }
+
+  hasNativeShapeShift(): boolean {
+    return false;
+  }
+
+  describePath(msg: core.DescribePath): core.PathDescription {
+    switch (msg.coin) {
+      case "Bitcoin":
```
kaladinlight, line 45: We need to add `BitcoinCash` here as well. I also changed the others to lowercase and added `toLowerCase()` so we don't have to remember casing and we aren't using enums atm

PR #167, opened 2020-07-13
Improve Native Wallet Initialization Performance

natven:
Derive seed once using bip39.js mnemonicToSeed. Use that seed to initialize each native wallet.

### DIFF
--- a/packages/hdwallet-native/src/ethereum.ts
+++ b/packages/hdwallet-native/src/ethereum.ts
@@ -1,7 +1,7 @@
 import * as core from "@shapeshiftoss/hdwallet-core";
 import { Wallet, utils } from "ethers";
 import txDecoder from "ethereum-tx-decoder";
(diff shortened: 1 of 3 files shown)


packages/hdwallet-native/src/ethereum.ts
```diff
@@ -43,8 +44,8 @@ export function MixinNativeETHWallet<TBase extends core.Constructor>(Base: TBase
 
     ethWallet: Wallet;
 
-    ethInitializeWallet(mnemonic: string): void {
-      this.ethWallet = Wallet.fromMnemonic(mnemonic);
+    ethInitializeWallet(seed: string): void {
+      this.ethWallet = new Wallet(HDNode.fromSeed(seed).derivePath(defaultPath));
```
elmutt, line 48: So fromMnemonic is just super slow for some reason?

natven: yep

PR #173, opened 2020-08-06
Feature eos native

BitHighlander:
(WIP)

# Bump reference tx to mainnet. Add EOS Asset and EOS network support to native

This pr Updates the reference EOS transaction example. It also enables signing of EOS tx's in native. 

Notes:
Reference tx can be verified on chain as follow

(This PR ONLY covers a transfer TX. Governance tx's will be a quick follow up)

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)


BitHighlander:
someday eos

PR #176, opened 2020-08-11
mobile friendly cosmos

BitHighlander:
This PR removes HDkey that is dependent on "crypto" lib only available in node.


packages/hdwallet-native/src/cosmos.ts
```diff
@@ -60,25 +62,26 @@ export function MixinNativeCosmosWallet<TBase extends core.Constructor>(Base: TB
...
+      return address;
     }
 
     async cosmosSignTx(msg: core.CosmosSignTx): Promise<core.CosmosSignedTx> {
       const seed = await mnemonicToSeed(this.#seed);
       const ATOM_CHAIN = "cosmoshub-3";
 
-      // expects bip32
+      const network = getNetwork("bitcoin");
+      const hdkey = bitcoin.bip32.fromSeed(seed, network);
```
kaladinlight, line 79: Update var name since we aren't using hdkey anymore

PR #179, opened 2020-08-13
Roll back Crypto-js to react-native support

BitHighlander:
Crypto-js  version 4.0 is broke. see https://github.com/parse-community/Parse-SDK-JS/issues/1201

This patch rolls back crypto-js and bumps version.


packages/hdwallet-native/package.json
```diff
@@ -14,19 +14,19 @@
...
   "dependencies": {
-    "@shapeshiftoss/hdwallet-core": "^1.1.1",
+    "@shapeshiftoss/hdwallet-core": "^1.1.2",
     "bchaddrjs": "^0.4.9",
     "bip32": "^2.0.5",
     "bip39": "^3.0.2",
     "bitcoinjs-lib": "^5.1.10",
     "cosmos-tx-builder": "^6.3.3",
-    "crypto-js": "^4.0.0",
+    "crypto-js": "^3.1.9-1",
```
BitHighlander, line 23: :magic:

majorhayes: What does this fix? I'm surprised the version of this matters because 1) we try to use native crypto functions in the mobile app, not JS ones, and 2) we have a tool that patches libraries to work with react native.

PR #183, opened 2020-08-31
clean up package.json

BitHighlander:
Bump support for node ver 14+

Add command for pre-reqs

Fix npm scripts

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,10 +21,11 @@
     ]
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -21,10 +21,11 @@
     ]
   },
   "scripts": {
+    "pre-reqs": "npm install -g typescript prettier",
```
kaladinlight, line 24: why global install of these packages? they are installed in the project.

PR #187, opened 2020-09-10
Add "MNEMONIC_REQUIRED" event for native wallet

majorhayes:
For the native wallet to work, it needs to be given the seed phrase. Similar to how KeepKey raises a PIN_REQUEST or PASSPHRASE_REQUEST event, this adds a new MNEMONIC_REQUIRED event that can be used to notify callers that the wallet is not initialized with a seed phrase.

* hdwallet-native needs to be able to be initialized without a mnemonic
* Raise MNEMONIC_REQUIRED whenever interacting with the seed
* Raise READY event after `loadDevice` is called to set the mnemonic


packages/hdwallet-native/src/native.ts
```diff
@@ -84,11 +85,17 @@ export class NativeHDWallet
...
   #mnemonic: string;
+  #transport: core.Transport;
 
-  constructor(mnemonic: string, deviceId: string) {
+  constructor(transport: core.Transport) {
     super();
-    this.#mnemonic = mnemonic;
-    this.#deviceId = deviceId;
+    this.#transport = transport;
+    // Prevent malicious JS from overwriting "loadDevice" to steal a mnemonic
```
elmutt, line 93: How would this happen?  Can you give an example?

examples/sandbox/index.ts
```diff
@@ -48,17 +48,18 @@ const keepkeyAdapter = WebUSBKeepKeyAdapter.useKeyring(keyring);
...
 
 const log = debug.default("hdwallet");
 
 keyring.onAny((name: string[], ...values: any[]) => {
+  console.log("Event:", name, values);
   const [[deviceId, event]] = values;
   const { from_wallet = false, message_type } = event;
   let direction = from_wallet ? "🔑" : "💻";
   debug.default(deviceId)(`${direction} ${message_type}`, event);
+  alert(`Event: ${name[2]}`);
```
kaladinlight, line 62: Probably want to remove this debugging?

examples/sandbox/index.ts
```diff
@@ -175,12 +176,6 @@ async function deviceConnected(deviceId) {
     console.error("Could not initialize PortisAdapter", e);
   }
 
-  try {
```
kaladinlight, line 178: Leave it auto initialize now that the event has been tested

package.json
```diff
@@ -25,8 +25,7 @@
     "clean": "rm -rf packages/*/tsconfig.tsbuildinfo && rm -rf coverage test-report && lerna clean --yes && rm -rf node_modules && yarn cache clean",
     "lint:ts": "tsc --noEmit",
     "format": "prettier .",
-    "yarn": "yarn",
-    "build": "yarn tsc --build",
+    "build": "lerna run build",
```
kaladinlight, line 28: This was working as intended using tsconfig paths/references
```suggestion
    "build": "yarn tsc --build",
```

package.json
```diff
@@ -25,8 +25,7 @@
     "clean": "rm -rf packages/*/tsconfig.tsbuildinfo && rm -rf coverage test-report && lerna clean --yes && rm -rf node_modules && yarn cache clean",
```
kaladinlight, line 25: and really this should probably look like
```suggestion
    "clean": "lerna run clean --scope @shapeshiftoss/* && rm -rf coverage test-report node_modules && yarn cache clean",
```

elmutt (approved):

packages/hdwallet-native/src/binance.ts
```diff
@@ -35,16 +37,57 @@ export function MixinNativeBinanceWalletInfo<TBase extends core.Constructor>(Bas
       // Only support one account for now (like portis).
       return undefined;
     }
+
+    binanceDescribePath(path: BIP32Path): PathDescription {
```
kaladinlight, line 41: Would like to see this in core/binance.ts

packages/hdwallet-native/src/cosmos.ts
```diff
@@ -34,10 +36,49 @@ export function MixinNativeCosmosWalletInfo<TBase extends core.Constructor>(Base
       // Only support one account for now (like portis).
       return undefined;
     }
+
+    cosmosDescribePath(path: BIP32Path): PathDescription {
```
kaladinlight, line 40: And this into `core/cosmos.ts`

packages/hdwallet-native/src/cosmos.ts
```diff
@@ -34,10 +36,49 @@ export function MixinNativeCosmosWalletInfo<TBase extends core.Constructor>(Base
...
+        coin: "Atom",
+        isKnown: true,
+        isPrefork: false,
+      };
+    }
   };
 }
 
-export function MixinNativeCosmosWallet<TBase extends core.Constructor>(Base: TBase) {
+export function MixinNativeCosmosWallet<TBase extends core.Constructor<NativeHDWalletBase>>(Base: TBase) {
```
kaladinlight, line 81: CosmosWallet needs the `needsMnemonic` wrapper where applicable

packages/hdwallet-native/src/ethereum.ts
```diff
@@ -38,46 +39,52 @@ export function MixinNativeETHWalletInfo<TBase extends core.Constructor>(Base: T
...
     ethInitializeWallet(seed: string): void {
-      this.#ethWallet = new Wallet(HDNode.fromSeed(seed).derivePath(defaultPath));
+      return this.needsMnemonic(!!seed, () => {
+        this.#ethWallet = new Wallet(HDNode.fromSeed(seed).derivePath(defaultPath));
+      });
     }
 
     async ethGetAddress(msg: core.ETHGetAddress): Promise<string> {
-      return this.#ethWallet.getAddress();
+      return this.needsMnemonicAsync(!!this.#ethWallet, () => this.#ethWallet.getAddress());
```
kaladinlight, line 55:
```suggestion
      return this.needsMnemonicAsync(!!this.#ethWallet, async () => this.#ethWallet.getAddress());
```

packages/hdwallet-native/src/adapter.ts
```diff
@@ -1,49 +1,44 @@
...
+    this.#mnemonic = args.mnemonic;
   }
 
   static useKeyring(keyring: core.Keyring, args: NativeAdapterArgs) {
     return new NativeAdapter(keyring, args);
   }
 
   async initialize(): Promise<number> {
-    let wallet = this.keyring.get(this.deviceId);
+    let wallet = this.keyring.get<NativeHDWallet>(this.deviceId);
```
kaladinlight, line 25: If we generic to extend `HDWallet` we should probably update the rest of the wallets as well? A side question would be is this necessary?

packages/hdwallet-native/src/native.ts
```diff
@@ -7,10 +9,57 @@ import { MixinNativeBTCWallet, MixinNativeBTCWalletInfo } from "./bitcoin";
...
+
+    this.#events.emit(
+      NativeEvents.MNEMONIC_REQUIRED,
+      core.makeEvent({
+        message_type: NativeEvents.MNEMONIC_REQUIRED,
+        from_wallet: true,
+      })
+    );
+
+    throw new Error("Wallet is not initialized");
```
kaladinlight, line 48: You mentioned the throw being problematic, should we just emit the event?

packages/hdwallet-native/src/native.ts
```diff
@@ -7,10 +9,57 @@ import { MixinNativeBTCWallet, MixinNativeBTCWalletInfo } from "./bitcoin";
...
+  }
+
+  async needsMnemonicAsync<T>(hasMnemonic: boolean, callback: () => Promise<T>): Promise<T | null> {
+    try {
+      return await this.needsMnemonic(hasMnemonic, callback);
```
kaladinlight, line 53: And then here I feel like we should maybe not wrap in a try to allow the callback function to throw as it would be intended to instead of potentially eating the error.

packages/hdwallet-native/src/native.ts
```diff
@@ -7,10 +9,57 @@ import { MixinNativeBTCWallet, MixinNativeBTCWalletInfo } from "./bitcoin";
...
+  }
+
+  get events() {
+    return this.#events;
+  }
+
+  /**
+   * Wrap a function call that needs a mnemonic seed
+   * @param hasMnemonic - A
+   * @param callback
```
kaladinlight, line 33: If we are going to doc comment the params, can we type it and provide a useful description if applicable?
`@param {boolean} hasMnemonic - useful description`

packages/hdwallet-native/src/native.ts
```diff
@@ -7,10 +9,57 @@ import { MixinNativeBTCWallet, MixinNativeBTCWalletInfo } from "./bitcoin";
...
+      core.makeEvent({
+        message_type: NativeEvents.MNEMONIC_REQUIRED,
+        from_wallet: true,
+      })
+    );
+
+    throw new Error("Wallet is not initialized");
+  }
+
+  async needsMnemonicAsync<T>(hasMnemonic: boolean, callback: () => Promise<T>): Promise<T | null> {
```
kaladinlight, line 51: And might as well document this one as well

packages/hdwallet-native/src/native.ts
```diff
@@ -7,10 +9,57 @@ import { MixinNativeBTCWallet, MixinNativeBTCWalletInfo } from "./bitcoin";
...
 import { MixinNativeBinanceWalletInfo, MixinNativeBinanceWallet } from "./binance";
+import type { NativeAdapterArgs } from "./adapter";
+
+export enum NativeEvents {
+  MNEMONIC_REQUIRED = "MNEMONIC_REQUIRED",
+  READY = "READY",
+}
+
+export class NativeHDWalletBase {
+  readonly #events: EventEmitter2;
```
kaladinlight, line 20: does this need to be a private field? not sure if it matters, but just curious as to your thought making it private.

kaladinlight:
Also remember to bump version and update changelog before merging

examples/sandbox/index.ts
majorhayes: Well, it's for the sandbox and I found it really helpful to show when an event was happening. I suppose I could have it just update some text on the page somewhere.

examples/sandbox/index.ts
majorhayes: I took this out when the initialization was throwing an error for not having a mnemonic. Since I took out the error, you're right that we can probably have it auto initialize now.

package.json
majorhayes: It was definitely NOT working for me. Nothing was building at all, that's why I changed it. I wouldn't have noticed if it was working.

package.json
majorhayes: I don't think I changed that line but I can add it.

packages/hdwallet-native/src/binance.ts
majorhayes: it's currently only implemented in native and keepkey so I wasn't sure if it was appropriate to have it in core if it doesn't apply to all wallets. What do you think?

packages/hdwallet-native/src/cosmos.ts
majorhayes: Can't believe I missed that. Thanks.

packages/hdwallet-native/src/ethereum.ts
majorhayes: The anonymous function doesn't need to be marked `async` because `getAddress` is async. That's why TypeScript isn't giving an error here because the return type is Promise.

packages/hdwallet-native/src/adapter.ts
majorhayes: The reason this is here is because the `events` property is unique to NativeHDWallet. If I leave it as a generic HDWallet, there's a compilation error for accessing a property that doesn't exist. `keyring.get` does have a type constraint that all generic types must extend HDWallet.

examples/sandbox/index.ts
kaladinlight: Yeah I would rather it not create an alert box on any event personally. Do you know if the `debug.default` logs out the events to console as is already?

packages/hdwallet-native/src/native.ts
majorhayes: there's only one place that uses the synchronous version of the function and I didn't notice any problems with that, but I didn't do a ton of testing. I'll take a look at this again.

package.json
kaladinlight: What were you seeing then? It is using the references to build packages in order based on any dependencies.

package.json
kaladinlight: I just noticed it when looking at the build script. Thanks

packages/hdwallet-native/src/native.ts
majorhayes: I'll take a look at this again. So far in my tests, any of these functions throwing an error was not being handled by axiom and was causing a crash, but I see your point about separating the error from `needsMnemonic` from the callback itself.

packages/hdwallet-native/src/native.ts
majorhayes: I didn't meant to leave that in.

packages/hdwallet-native/src/binance.ts
kaladinlight: That was the pattern I moved towards for native and think it makes more sense since it should be the same for all wallets. Also if we add to ledger or trezor, we won't need to copy it there as well. Would be my preference.

packages/hdwallet-native/src/native.ts
majorhayes: I wanted it to be read-only at runtime. Probably not a big deal either way.

majorhayes:
> Also remember to bump version and update changelog before merging

About that... I'm not sure that it's the right thing to do to do a version bump in a feature PR. I think it's better to have a separate version bump specifically when we're ready for a release to go out. It's not uncommon for two PRs to be open at the time and if both have version bumps you have to resolve conflicts depending on which gets merged first.

I would think that it would be better to just merge the PR and then do a release PR to bump the version.

We have the same issue with axiom packages.

packages/hdwallet-native/src/ethereum.ts
kaladinlight: Yeah that makes sense. It "looks" more right in ref to `needsMnemonicAsync` without having to know `getAddress` return type. I guess like you said typescript would complain if need be. Nitpick for looks overall. Your call.

kaladinlight:
> > Also remember to bump version and update changelog before merging
> 
> About that... I'm not sure that it's the right thing to do to do a version bump in a feature PR. I think it's better to have a separate version bump specifically when we're ready for a release to go out. It's not uncommon for two PRs to be open at the time and if both have version bumps you have to resolve conflicts depending on which gets merged first.
> 
> I would think that it would be better to just merge the PR and then do a release PR to bump the version.
> 
> We have the same issue with axiom packages.

I can get behind that for sure. Just want to make sure we don't forget it as I have done that before haha

packages/hdwallet-native/src/native.ts
kaladinlight: Yeah, we should probably keep track of where axiom doesn't handle hdwallet errors correctly and actually address them

packages/hdwallet-native/src/native.ts
kaladinlight: ok, just curious more than anything

examples/sandbox/index.ts
majorhayes: I went ahead and put some time into implementing a mnemonic_required modal in sandbox and added an "event log" section rather than relying on "alert".

examples/sandbox/index.ts
majorhayes: I put it back in, but I had to remove a `wallet.initialize()` line in the "on("change")` handler for the select at the top to avoid always popping up the mnemonic required modal.

package.json
```diff
@@ -22,15 +22,15 @@
...
+    "clean": "lerna run clean --scope @shapeshiftoss/* && rm -rf coverage test-report node_modules && yarn cache clean",
     "lint:ts": "tsc --noEmit",
     "format": "prettier .",
-    "yarn": "yarn",
     "build": "yarn tsc --build",
     "dev:all": "yarn tsc --build --watch",
     "now-build": "lerna run build --scope @shapeshiftoss/hdwallet-core && lerna run build --scope hdwallet-sandbox --ignore @shapeshiftoss/hdwallet-keepkey-nodehid --include-filtered-dependencies && mv examples/sandbox/public .",
     "docker:run:emulator": "docker run -p 5000:5000 kktech/kkemu:latest",
     "docker:stop:run:emulator": "docker stop $(docker ps -a -q --filter ancestor=kktech/kkemu:latest --format=\"{{.ID}}\") && docker run -p 5000:5000 kktech/kkemu:latest",
+    "sandbox": "cd examples/sandbox && yarn dev",
```
kaladinlight, line 33: so this actually used to `dev:sandbox` I had accidentally removed it a little while back and it is still referenced in the readme. can we use this instead?
```suggestion
    "dev:sandbox": "lerna run dev --scope hdwallet-sandbox --parallel --include-filtered-dependencies",
```

packages/hdwallet-native/src/native.ts
```diff
@@ -82,15 +125,20 @@ export class NativeHDWallet
   _isNative = true;
 
   readonly #deviceId: string;
+  transport?: core.Transport;
```
kaladinlight, line 128: Don't need this anymore

packages/hdwallet-native/src/native.ts
```diff
@@ -55,12 +94,16 @@ class NativeHDWalletInfo
       case "testnet":
         const unknown = core.unknownUTXOPath(msg.path, msg.coin, msg.scriptType);
 
-        if (!super.btcSupportsCoin(msg.coin)) return unknown;
-        if (!super.btcSupportsScriptType(msg.coin, msg.scriptType)) return unknown;
+        if (!this.btcSupportsCoin(msg.coin)) return unknown;
+        if (!this.btcSupportsScriptType(msg.coin, msg.scriptType)) return unknown;
```
kaladinlight, line 98: Can this stay as super to show we are calling a function from an extended class.

packages/hdwallet-native/src/adapter.ts
```diff
@@ -1,49 +1,48 @@
...
 
     if (!wallet) {
-      wallet = create(this.mnemonic, this.deviceId);
+      wallet = create({ mnemonic: this.#mnemonic, deviceId: this.deviceId });
       this.keyring.add(wallet, this.deviceId);
+      this.keyring.decorateEvents(this.deviceId, wallet.events);
     }
 
-    wallet.initialize();
+    if (this.#mnemonic) {
```
kaladinlight, line 33: We don't really need to wrap this is the mnemonic check now with the addition of `needsMnemonic` in `wallet.initialize`. Thoughts?

packages/hdwallet-native/src/native.ts
majorhayes: my IDE automatically added it I guess because it's technically part of the interface, but I'll remove it.

packages/hdwallet-native/src/native.ts
majorhayes: I read up on this before the change, and unless overridden, they are basically the same thing. Since it's the child, `this` works just fine. It's not really worth arguing about though.

packages/hdwallet-native/src/adapter.ts
majorhayes: that's actually WHY we need it, because otherwise it'll immediate raise the mnemonic required event immediately on initializing the adapter, which is what I didn't want to happen.

majorhayes: BTW, in my head I'm considering "adapter.initialize()" as a separate function from "wallet.initialize". I see the former as setting up the adapter and the latter as setting up the wallet. In our case, setting up the adapter itself shouldn't require the mnemonic, only initializing the wallet itself should. Maybe I'm thinking about that wrong, or not how it was intended?

kaladinlight: No, you are good, I confused myself. You are correct

packages/hdwallet-native/src/native.ts
kaladinlight: Sure, I would just say we are using super to call all parent class function except for this function now is all. I will leave it at that.

kaladinlight (approved):

PR #191, opened 2020-09-18
Move common helper functions into core

majorhayes:
Every wallet has copy/paste copies of the same coin helper functions.  This PR moves some of those helper functions into core and changes each wallet to just reference the version of the function in core.

I made an effort to compare the function in each wallet to uniqueness before pulling it out into core. All of these functions were exact duplicates, but it may be worth doing a quick double-check to see if you see anything different in any of the removed functions.

Prettier formatting was automatically applied on save, so there are some formatting changes.


majorhayes:
So, it turns out that there are enough very subtle differences between the various wallets, that consolidating this caused some problems and not worth the work right now.

PR #193, opened 2020-09-21
Support native binanceSignTx again.

kaladinlight:


packages/hdwallet-native/src/binance.ts
```diff
@@ -42,14 +41,24 @@ export function MixinNativeBinanceWalletInfo<TBase extends core.Constructor>(Bas
 export function MixinNativeBinanceWallet<TBase extends core.Constructor<NativeHDWalletBase>>(Base: TBase) {
   return class MixinNativeBinanceWallet extends Base {
     _supportsBinance = true;
-    #seed: Buffer;
+
+    #binanceSeed: Buffer;
 
     async binanceInitializeWallet(mnemonic: string): Promise<void> {
-      this.#seed = await mnemonicToSeed(mnemonic);
+      this.#binanceSeed = await mnemonicToSeed(mnemonic);
```
BitHighlander, line 48: this feels weird

BitHighlander (approved):

packages/hdwallet-native/src/binance.ts
kaladinlight: Made the variables explicit as the coin wallets have been set up to interact with them as distinct classes and with the mixin inheritance model, anything that shares the same name will be overridden by the last subclass.

PR #201, opened 2020-10-02
text encode differently based on environment

elmutt:
text encode differently based on environment

formatting bs


majorhayes:
Is there a TextDecoder polyfill available that we can use that doesn't require us to use our own logic to switch between them? For example, we're using Intl.NumberFormat polyfill that includes a check on whether or not it needs to be loaded and them we just have to import `Intl` without checking if the polyfill is required or not.

BitHighlander (approved):

elmutt:
> Is there a TextDecoder polyfill available that we can use that doesn't require us to use our own logic to switch between them? For example, we're using Intl.NumberFormat polyfill that includes a check on whether or not it needs to be loaded and them we just have to import `Intl` without checking if the polyfill is required or not.

The fio docs didnt have a lot of info but basically said in node use one and in web use the other.  I struggled for a while trying to figure out why the encrypt/derypt was failing appearign to succeed but missing information in the fields.  Finally I found a note about it in their readme.  I understand what you are saying but hiven the aggressive timeline for fio this was good enough for me

PR #204, opened 2020-10-05
Refactor public "getKeyPair" function to be private

majorhayes:
binance and cosmos support in hdwallet-native exposed a public `getKeyPair` function, which was used as a helper function but was also a class method, making it public. Malicious injected JavaScript would be able to call `wallet.binanceGetKeyPair` to get the PRIVATE key from the seed.

This refactors the code so the getKeyPair is a static method that requires providing the seed as a parameter. Thus, the private nature of the `#wallet` private variable is maintained.

Because of the nature of the MixIn class, it's not possible to make a private instance method.


BitHighlander (approved):
:shipit:

PR #213, opened 2020-10-14
hdwallet-native: Add crypto code to be able to create and decrypt wallets

majorhayes:
# EncryptedWallet
* Add an `EncryptedWallet` class
  * Uses an email and password to generate an encryption key
  * Generates a password hash for authentication so the password isn't exposed to servers
  * Generates an encrypted wallet string that can later be decrypted using the same email and password
  * Can decrypt an encrypted wallet

## Testing
Automated testing wasn't working reliability, and automated tests were crucial to making sure the `EncryptedWallet` class is correct.

* Moved `jest`, `ts-jest`, and `@types` dev dependencies into the root workspace
* Updated `jest.config.js` to only run unit tests (integration tests will need to be run separately)


packages/hdwallet-native/src/crypto/CryptoHelper.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  async makeKey(password: string, email: string): Promise<SymmetricCryptoKey> {
+    if (!password || !email) {
+      throw new Error("A password and email are required to make a symmetric crypto key.");
+    }
+
+    const salt = utils.toArrayBuffer(email);
+    const key = await this.#engine.scrypt(utils.toArrayBuffer(password), salt, {
+      iterations: 16384,
```
elmutt, line 119: magic numbers?

packages/hdwallet-native/src/crypto/classes/symmetricCryptoKey.test.ts
```diff
@@ -0,0 +1,45 @@
+import { EncryptionType } from "./encryptionType";
+import { SymmetricCryptoKey } from "./symmetricCryptoKey";
+
+describe("symmetricCryptoKey", () => {
+  const key = new Uint8Array(32).fill(64);
```
elmutt, line 5: is this filling the array with a specific bit pattern?  Would a hexadecimal representation be better?

elmutt (approved):

packages/hdwallet-native/src/crypto/classes/symmetricCryptoKey.test.ts
majorhayes: this is just for the test, the contents of the key are irrelevant for the test. This is creating a 32-byte array filled with the byte 64 32 times.

packages/hdwallet-native/src/crypto/CryptoHelpers.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      const key = await helper.makeKey("password", "email");
+      const encrypted = await helper.aesEncrypt(toArrayBuffer("totally random secret data"), key);
+      expect(encrypted.key).toEqual(key);
+      expect(encrypted.iv.byteLength).toBe(16);
+      expect(encrypted.mac.byteLength).toBe(32);
+      expect(encrypted.data.byteLength).toBe(32);
+    });
+  });
+
+  describe("aesDecrypt", () => {});
```
millsmcilroy, line 37: ?

packages/hdwallet-native/src/crypto/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!(password && typeof password === "string" && password.length > 0)) {
+      throw new Error("Invalid password");
+    }
+
+    this.#email = email.normalize("NFKC").trim().toLowerCase();
+    this.#password = password.normalize("NFKC");
+    this.#key = await this.#helper.makeKey(this.#password, this.#email);
+
+    if (encryptedWallet) {
+      this.encryptedWallet = encryptedWallet;
```
millsmcilroy, line 61: Why not `this.#encryptedWallet` here?

packages/hdwallet-native/src/crypto/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    return this;
+  }
+
+  async createWallet() {
+    if (!this.isInitialized) throw new Error("Wallet is not initialized");
```
millsmcilroy, line 68: `isInitialized` is a function, needs to be called like this `this.isInitialized()`

packages/hdwallet-native/src/crypto/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      throw new Error("Invalid mnemonic");
+    }
+
+    this.#encryptedWallet = (await this.#helper.aesEncrypt(utils.toArrayBuffer(mnemonic), this.#key)).toString();
+
+    return this;
+  }
+
+  async decrypt(encryptedWallet = this.#encryptedWallet): Promise<string> {
+    if (!this.isInitialized) throw new Error("Wallet is not initialized");
```
millsmcilroy, line 81: Same as above

packages/hdwallet-native/src/crypto/CryptoHelpers.test.ts
majorhayes: looks like I forgot to add some tests.

packages/hdwallet-native/src/crypto/index.ts
majorhayes: because the `set encryptedWallet` function runs it through `CipherString` which does some validation, and it provides consistency for how that variable is set.

packages/hdwallet-native/src/crypto/index.ts
majorhayes: its a getter, so no.

PR #216, opened 2020-10-16
native: Speed up initialization by running all coins in parallel

majorhayes:
* Since all coins convert mnemonic to seed, just do it once and pass the seed buffer to each init function
* fio: SDK only exposes using a mnemonic, so I had to copy over the relevant code

### DIFF
--- a/packages/hdwallet-native/src/native.ts
+++ b/packages/hdwallet-native/src/native.ts
@@ -196,11 +196,15 @@ export class NativeHDWallet
   async initialize(): Promise<boolean> {
     return this.needsMnemonic(!!this.#mnemonic, async () => {
       try {
-        await super.btcInitializeWallet(this.#mnemonic);
-        await super.ethInitializeWallet(this.#mnemonic);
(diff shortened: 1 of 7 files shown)


BitHighlander (approved):
:shipit:

packages/hdwallet-native/src/native.ts
```diff
@@ -196,11 +196,15 @@ export class NativeHDWallet
...
     return this.needsMnemonic(!!this.#mnemonic, async () => {
       try {
-        await super.btcInitializeWallet(this.#mnemonic);
-        await super.ethInitializeWallet(this.#mnemonic);
-        await super.cosmosInitializeWallet(this.#mnemonic);
-        await super.binanceInitializeWallet(this.#mnemonic);
-        await super.fioInitializeWallet(this.#mnemonic);
+        const seed = await mnemonicToSeed(this.#mnemonic);
+
+        await Promise.all([
```
millsmcilroy, line 201: 💯

PR #219, opened 2020-10-19
Try to fix master builds by resetting yarn cache

majorhayes:
Builds on PRs are successful, but "master" builds are failing do to what looks like are conflicts between two yarn caches.

This changes the yarn cache key to attempt to start from a clean cache and avoid the build error.


BitHighlander (approved):
:shipit:

PR #221, opened 2020-10-20
native: Add CryptoEngine requirement for SHA hashing

majorhayes:
* Add `deviceId` property to the wallet - this is so we can use the same ID across platforms
* Add more tests
* Export the `EncryptedWallet` and `WebCryptoEngine` classes - these weren't properly exported in the previous release


packages/hdwallet-native/src/crypto/engines/web-crypto.ts
```diff
@@ -1,15 +1,20 @@
...
-    return await globalThis.crypto.subtle.decrypt({ name: "AES-CBC", iv }, impKey, data);
+    return globalThis.crypto.subtle.decrypt({ name: "AES-CBC", iv }, impKey, data);
+  }
+
+  public async digest(algorithm: DigestAlgorithm, data: ArrayBuffer): Promise<ArrayBuffer> {
+    const alg = algorithm === DigestAlgorithm.SHA512 ? "SHA-512" : "SHA-256";
+    return globalThis.crypto.subtle.digest(alg, data);
   }
 
   public async encrypt(data: ArrayBuffer, key: ArrayBuffer, iv: ArrayBuffer): Promise<ArrayBuffer> {
```
elmutt, line 15: does this still need to be async?

packages/hdwallet-native/src/crypto/engines/web-crypto.ts
```diff
@@ -1,15 +1,20 @@
...
 
-export default class WebCryptoEngine implements CryptoEngine {
+export class WebCryptoEngine implements CryptoEngine {
   public async decrypt(data: ArrayBuffer, key: ArrayBuffer, iv: ArrayBuffer): Promise<ArrayBuffer> {
     const impKey = await globalThis.crypto.subtle.importKey("raw", key, { name: "AES-CBC" }, false, ["decrypt"]);
-    return await globalThis.crypto.subtle.decrypt({ name: "AES-CBC", iv }, impKey, data);
+    return globalThis.crypto.subtle.decrypt({ name: "AES-CBC", iv }, impKey, data);
+  }
+
+  public async digest(algorithm: DigestAlgorithm, data: ArrayBuffer): Promise<ArrayBuffer> {
```
elmutt, line 10: this need to be async?

packages/hdwallet-native/src/crypto/CryptoHelper.ts
```diff
@@ -148,4 +151,18 @@ export default class CryptoHelper {
     const entropy = await this.#engine.randomBytes(strength / 8);
     return entropyToMnemonic(Buffer.from(entropy));
   }
+
+  /**
+   * Generates a base64 hash based on the provided data
+   * Should be used to calculate a device ID for the wallet based on the mnemonic
+   */
+  async getDeviceId(data: string): Promise<string> {
+    if (!(typeof data === "string" && data.length > 0)) throw new Error("Invalid data to hash");
```
elmutt, line 160: The ! out front makes the logic harder for me to read. I know it doesnt really matter but  what about "dataType != string || data.length <= 0" or something?

elmutt (approved):

packages/hdwallet-native/src/crypto/CryptoHelper.ts
majorhayes: I do it like that because it's easier for me to think of the cases where the validation passes than when it doesn't, so I write the good cases then wrap that in the `!` to negate it, but I do understand where you're coming from.
I'd prefer to use a standard validation process (library or home grown) to make everything consistent.

packages/hdwallet-native/src/crypto/engines/web-crypto.ts
majorhayes: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

> digest is a Promise that fulfills with an ArrayBuffer containing the digest.

majorhayes: Also, there is a parent contract defined called CryptoEngine in which all the functions are declared `async` because it's easier to return a synchronous result as a Promise than the other way around. Currently, both the web crypto and React Native crypto use async functions for all crypto code.

majorhayes:
The big change for `EncryptedWallet` is actually just a rename of the file to facilitate exporting easier, it's not actually a whole new file.

packages/hdwallet-native/src/crypto/CryptoHelper.ts
millsmcilroy: Do you need the `typeof data === "string"` check if the function signature requires data to be a string?

packages/hdwallet-native/src/crypto/EncryptedWallet.test.ts
```diff
@@ -120,10 +120,42 @@ describe("EncryptedWallet", () => {
       await expect(result.decrypt()).resolves.toEqual(PLAIN);
     });
 
-    it("should require the wallet to be initialized", async () => {
+    it("should require that the wallet to be initialized", async () => {
```
millsmcilroy, line 123: `that the wallet to be initialized` doesn't sound right

packages/hdwallet-native/src/crypto/EncryptedWallet.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (!(password && typeof password === "string" && password.length > 0)) {
+      throw new Error("Invalid password");
+    }
+
+    this.#email = email.normalize("NFKC").trim().toLowerCase();
+    this.#password = password.normalize("NFKC");
+    this.#key = await this.#helper.makeKey(this.#password, this.#email);
+
+    if (encryptedWallet) {
+      this.encryptedWallet = encryptedWallet;
```
millsmcilroy, line 86: This assignment could be done in the constructor right?

millsmcilroy (approved):
LGTM. I left a couple questions.

packages/hdwallet-native/src/crypto/EncryptedWallet.test.ts
majorhayes: ah damn.

packages/hdwallet-native/src/crypto/EncryptedWallet.ts
majorhayes: just the encrypted wallet part? Right now the constructor only takes in the engine then requires an init call. `encryptedWallet` by itself isn't useful because you need the email/password to do anything with it anyway

packages/hdwallet-native/src/crypto/CryptoHelper.ts
majorhayes: TypeScript only enforces types at compile time not at run time. This is an open source library and anyone could use it with plain JavaScript. There's nothing in that scenario that will prevent them from passing invalid data. I prefer that libraries I use throw errors when I use them incorrectly rather than just do unexpected things that are hard to debug.

PR #226, opened 2020-10-23
chore(docs): add image of architecture

schnogz:
Took me forever to convert the `.dot` file to an image.  Hoping this saves people time in the future.


BitHighlander (approved):
😎

PR #227, opened 2020-10-27
native: Change adapter to support connecting more than one wallet

majorhayes:
Previously, the "deviceId" was kept at the adapter so only one wallet could ever be used.

* NativeAdapter no longer requires a deviceId to be instantiated
* `initialize` doesn't do anything as there's nothing to do. It doesn't create a default wallet like it did before.
* `pairDevice` now REQUIRES that you provide the `deviceId` upon pairing

This allows us fetch a list of wallets from a database, and create a separate `NativeHDWallet` for each one based on a unique ID.

### DIFF
--- a/packages/hdwallet-native/src/adapter.ts
+++ b/packages/hdwallet-native/src/adapter.ts
@@ -1,5 +1,5 @@
 import * as core from "@shapeshiftoss/hdwallet-core";
-import { create, NativeHDWallet } from "./native";
+import { NativeHDWallet } from "./native";
 
 export type NativeAdapterArgs = {
   mnemonic?: string;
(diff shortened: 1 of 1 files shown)


SteveHadow (commented):
LGTM but how have you been testing? playground?

Looks like there is at least one place in axiom web `WalletContext` that we will have to update to pass the deviceId through on `pairDevice` if the device is native. There should probably be a PR for that before we bump this version and release it.

millsmcilroy:
> Looks like there is at least one place in axiom web `WalletContext` that we will have to update to pass the deviceId through on `pairDevice` if the device is native. There should probably be a PR for that before we bump this version and release it.

Agreed. Since this is OSS we have to be mindful of anyone else who might be using it and do a major version bump if there are breaking changes.

majorhayes:
> > Looks like there is at least one place in axiom web `WalletContext` that we will have to update to pass the deviceId through on `pairDevice` if the device is native. There should probably be a PR for that before we bump this version and release it.
> 
> Agreed. Since this is OSS we have to be mindful of anyone else who might be using it and do a major version bump if there are breaking changes.

This PR doesn't change the version. That would be a separate PR after this is merged.

BitHighlander (approved):
:shipit:

majorhayes:
> LGTM but how have you been testing? playground?
> 
> Looks like there is at least one place in axiom web `WalletContext` that we will have to update to pass the deviceId through on `pairDevice` if the device is native. There should probably be a PR for that before we bump this version and release it.

The code for that is already done. I'm just waiting for the package to be deployed.

PR #234, opened 2020-12-10
Users can not send BNB in platform

mcchadwick:
Moved type enforcement for transaction amounts fields from Axiom to hdwallet-keepkey.


majorhayes:
there are other binance files in here, specifically for native. Can you review those as  well?

BitHighlander (dismissed):

BitHighlander (approved):
:shipit:

PR #241, opened 2021-02-18
restore from seed

BitHighlander:
option feature to restore with seed

### DIFF
--- a/packages/hdwallet-native/src/crypto/EncryptedWallet.ts
+++ b/packages/hdwallet-native/src/crypto/EncryptedWallet.ts
@@ -92,7 +92,7 @@ export class EncryptedWallet {
   /**
    * Generate a new mnemonic and encrypt it with the email and password
    */
-  async createWallet(mnemonic?) {
+  async createWallet(mnemonic?: string) {
     if (!this.isInitialized) throw new Error("Wallet is not initialized");
     if(!mnemonic) mnemonic = await this.#helper.generateMnemonic();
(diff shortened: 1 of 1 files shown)


packages/hdwallet-native/src/crypto/EncryptedWallet.ts
```diff
@@ -92,9 +92,9 @@ export class EncryptedWallet {
   /**
    * Generate a new mnemonic and encrypt it with the email and password
    */
-  async createWallet() {
+  async createWallet(mnemonic?) {
```
majorhayes, line 95: I would recommend a separate `importMnemonic` or `createWalletFromMnemonic` function.

The correct TypeScript type here would be `async createWallet(mnemonic?: string)`

packages/hdwallet-native/src/crypto/EncryptedWallet.ts
```diff
@@ -92,9 +92,9 @@ export class EncryptedWallet {
...
     if (!this.isInitialized) throw new Error("Wallet is not initialized");
-    const mnemonic = await this.#helper.generateMnemonic();
+    if(!mnemonic) mnemonic = await this.#helper.generateMnemonic();
```
majorhayes, line 97:
```suggestion
    mnemonic = mnemonic ?? await this.#helper.generateMnemonic();
```

mvpratt (approved):

PR #247, opened 2021-03-03
Optional btc op return data

elmutt:
Optional btc op return data


BitHighlander (approved):

packages/hdwallet-core/src/bitcoin.ts
```diff
@@ -89,6 +89,7 @@ export interface BTCSignTx {
   outputs: Array<BTCSignTxOutput>;
   version?: number;
   locktime?: number;
+  opReturnData: string;
```
kaladinlight, line 92: Make optional?
```suggestion
  opReturnData?: string;
```

kaladinlight (approved):

mcchadwick (approved):
I second Kevin’s suggestion about making the opReturnData field optional, but other than that, the P.R. looks good to me.

PR #250, opened 2021-03-23
Improved Tests Phase 1: Housekeeping

reidrankin:
This is the first (edit: yeah, that aged well) of a series of PRs targeting the `hdwallet-native` package. My end goal is to land net-new isolated key storage machinery; first, though, the tests need to be improved so that there's a baseline for evaluating whether the new stuff will break anything.

I'm planning a three-phase endeavor to achieve this:

1) Housekeeping: a few changes to existing code, small and manually-inspectable (this PR)
2) Test Improvements: new and improved tests, with no changes to non-test code or behavior
3) Bugfixes: changes to non-test code to fix the bugs exposed by the new tests

Some of the new tests added in phase 2 will fail because they expose underlying bugs. An alternative approach would be to introduce bugfixes and tests at the same time, so that the tests always pass; however, because of the sheer volume of new tests being added it's also valuable to be able to look at them as a unit and conclude that they can't break anything because they don't change any non-test behavior.

This first PR covers some simple, non-controversial housekeeping items (formatting, typos, a couple of logic bugs, and environment setup) in order to cleanly separate these concerns from the tests themselves.


reidrankin:
@millsmcilroy @natven

packages/hdwallet-native/jest.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  preset: "ts-jest",
+//   testEnvironment: "node",
+  testPathIgnorePatterns: ["dist"],
+  globals: {
```
reidrankin, line 5: Using a jest.config.js instead of embedding the configuration into package.json is required to pass through Uint8Array and ArrayBuffer globals, which we need to work around facebook/jest#4422.

packages/hdwallet-native/src/native.test.ts
```diff
@@ -1,10 +1,11 @@
-import { NativeHDWallet } from "./native";
+import * as NativeHDWallet from "./native";
```
reidrankin, line 1: Using `.create()` instead of the constructor 1) lets us make this import change, which will make things prettier later, and 2) tests a slightly-longer code path.

reidrankin:
(Oh yeah, just in case I wasn't clear: all this code is ready to go. The phased PR approach is just a process thing.)

reidrankin:
Converted to draft for a bit; want to make one more pass and make sure all the stuff that should be in here is in here

packages/hdwallet-native/src/binance.ts
```diff
@@ -79,20 +79,22 @@ export function MixinNativeBinanceWallet<TBase extends core.Constructor<NativeHD
...
         await client.chooseNetwork("mainnet");
-        await client.setPrivateKey(privateKey, Number.isInteger(Number(msg.account_number)));
+        const haveAccountNumber = Number.isInteger(Number(msg.account_number));
+        if (haveAccountNumber) await client.setAccountNumber(Number(msg.account_number));
+        await client.setPrivateKey(privateKey, haveAccountNumber);
         await client.initChain();
 
-        const addressFrom = msg.tx?.msgs?.[0]?.inputs?.[0]?.address;
-        const addressFromVerify = client.getClientKeyAddress()
+        const addressFrom = msg.tx.msgs[0].inputs[0].address;
```
reidrankin, line 87: The type specification on `msg` requires that this path exist, and if it doesn't a hard-fail is preferable to feeding `undefined` to the Binance SDK.

packages/hdwallet-native/package.json
```diff
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "@shapeshiftoss/hdwallet-core": "^1.11.1",
+    "@zxing/text-encoding": "^0.9.0",
```
reidrankin, line 18: This is an optional dependency of `web-encoding`, and provides `TextEncoder` support when it's not available natively. It's a fork of the now-unmaintained `text-encoding` package.

packages/hdwallet-native/src/crypto/classes/cipherString.ts
```diff
@@ -44,4 +45,13 @@ export class CipherString {
   get encryptedString() {
     return `${this.encryptionType}.${[this.data, this.iv, this.mac || ""].join("|")}`;
   }
+
+  toEncryptedObject(key: SymmetricCryptoKey): EncryptedObject {
```
reidrankin, line 49: This helper makes it easier to actually decrypt a CipherString, and will be used in tests later.

reidrankin:
It's apparent that the binance integration tests need a little TLC before this can land. Sorry for the premature tagging; please bear with me.

integration/src/binance/tx02.mainnet.signed.json
```diff
@@ -36,7 +36,7 @@
     }
   ],
   "txid": "A640902638C0F08B6BF225D18AE9A15FD0B6DBA6C739BD3F5E390CE737BC8E75",
-  "serialized": "c001f0625dee0a482a2c87fa0a200a14ea5d7ae99a9ce2f9db7c43a907377024b7f343e812080a03424e4210e80712200a14679cd81fab41426840fa60280970513b810eb41e12080a03424e4210e80712700a26eb5ae9872102ddf08b1f51a4132c63c73455cac6569404c78c9e3b5bffd1f64e07a9a2f109b412400e7eb2814f3659e49ff228f39d968e05d1c05894e412bb193e575ca2fcf109a84ec1df6ece4b85586717e53066baa9b3f9da6ccea07f4fd92e1046f86ca64fe818c9e01c2008",
+  "serialized": "c001f0625dee0a482a2c87fa0a200a14ea5d7ae99a9ce2f9db7c43a907377024b7f343e812080a03424e4210e80712200a14679cd81fab41426840fa60280970513b810eb41e12080a03424e4210e80712700a26eb5ae9872102ddf08b1f51a4132c63c73455cac6569404c78c9e3b5bffd1f64e07a9a2f109b412406cca9c5ac3d068d2acb58a8ff3ea68007120921eaecc7f4a2c31cfde9b753d4c158cae5719a03acf6524da57ca8d91634903c6dac183f789f088385f7784c78e18c9e01c2011",
```
reidrankin, line 35: Because tests pass before and after this commit, and the signature value is not changing, this updated value must be correct.

integration/src/binance/binance.ts
```diff
@@ -65,18 +68,63 @@ export function binanceTests(get: () => { wallet: HDWallet; info: HDWalletInfo }
           tx: tx02_unsigned,
           addressNList: bip32ToAddressNList("m/44'/714'/0'/0/0"),
           chain_id: "Binance-Chain-Nile",
-          account_number: "24250",
-          sequence: 0,
```
reidrankin, line 69: Note that these values were previously being ignored due to a bug in hdwallet-native.

integration/src/binance/tx02.mainnet.signed.json
```diff
@@ -1,40 +1,36 @@
...
-            "type": "Buffer",
-            "data": [234, 93, 122, 233, 154, 156, 226, 249, 219, 124, 67, 169, 7, 55, 112, 36, 183, 243, 67, 232]
-          },
+          "address": "bnb1afwh46v6nn30nkmugw5swdmsyjmlxslgjfugre",
           "coins": [
             {
-              "denom": "BNB",
-              "amount": "1000"
+              "amount": "0.00001",
+              "denom": "BNB"
```
reidrankin, line 14: The changes in field order and data type bring the test vectors in line with the `signBytes` value -- the serialized data that's actually signed. (Not to be confused with the serialized transaction, which is different for reasons.)

integration/src/binance/tx02.mainnet.signed.json
```diff
@@ -1,40 +1,36 @@
...
             }
           ]
         }
-      ],
-      "msgType": "MsgSend"
+      ]
     }
   ],
+  "sequence": "17",
+  "source": "0",
```
reidrankin, line 33: These values of sequence and data type match the ones actually signed -- which do not, until 2cb9480, actually come from here, but from an call out to the interwebs to check actual on-chain state of this address.

elmutt (approved):

PR #251, opened 2021-03-24
Improved Tests Phase 0: Get BNB Integration Tests Fixed (Part 1)

### LINKED ISSUE #266: BNB integration tests should ensure the serialized TX is as expected
Currently, the Binance integration tests have an incorrect test vector for the serialized TX, but still pass because they don't actually check it. The test vector should be updated to match the actual test transaction, and the tests should be updated to verify it.

reidrankin:
Fixes #266. Because the signatures' test vectors are not changed here, and the tests ensure those same signatures are being produced, you can be confident that this new serialized transaction test vector matches what was actually previously being signed.


SteveHadow (approved):

reidrankin:
Just realized this needs a patch-level version bump as it changes non-test behavior.

elmutt (approved):

kaladinlight (approved):

PR #265, opened 2021-04-01
Improved Tests Phase -1: Fix hdwallet-native BNB math

### LINKED ISSUE #263: hdwallet-keepkey should fail hard with invalid "amount" inputs
When processing BNB transactions, `hdwallet-keepkey` does not perform input validation before handing the "amount" value off to the protobuf library for serialization, which conveniently provides a textbook example of the sheer eldritch horror lurking behind the words "undefined behavior".

In particular, despite the applicable field in the structure being defined as an integer, the [upstream](https://github.com/protocolbuffers/protobuf/blob/aedb8d726c8d2a371d999354d44ca7e1bdbb1e73/js/binary/utils.js#L631) protobuf library doesn't perform any input checking. It happily parses the provided string character-by-character, subtracting the ascii value of `0` to get the numeric value of each digit. When it encounters `.`,  which has a *lower* value than zero, the fact that the subtraction results in a negative number doesn't stop it from computing with it, coercing the result to a hilariously-large unsigned integer, sticking that in an array that's *supposed* to only contain byte values, and then doing more non-type-checked serialization on *that*.

The design principal of speed-over-correctness seems to be a conscious choice by the authors of the upstream code, so the appropriate remedy would be improved type-checking in `hdwallet-keepkey`.

### LINKED ISSUE #264: hdwallet-native doesn't handle the amount field properly
`hdwallet` is supposed to take integer amounts of the smallest-denomination unit of each coin (satoshis for BTC, wei for ETH, etc). The Binance on-chain protocol uses this representation as well, but the BNB JS SDK, despite having a very-similar API surface, expects a decimal amount of BNB and "helpfully" converts it by multiplying by 10^8. This results in signing transactions for 10^8 too much value, which only safe if the resulting TX is invalid because you don't have that much to spend.

reidrankin:
Fixes #264 and then fixes #263.

This fix normalizes the behavior of BNB in hdwallet-native and BNB in everything else. The inputs to the BNB integration tests have been updated to reflect the correct input type, and are now consistent with the expected integration test outputs. The same string of bytes is still generated and signed, as evidenced by the updated inputs still producing the same signature value.

This change in integration test inputs also conveniently sidesteps #263. Because of this issue, the previous `kksignature` was made over an incorrect byte string; the signature is now made over the correct bytes, so the expected output had to be updated to match.

Per @natven this will require a minor version bump, which I'm pretty sure requires someone with write access to run the appropriate `lerna publish` command after this is merged to set up the tags and such.

### DIFF
--- a/packages/hdwallet-keepkey/src/binance.ts
+++ b/packages/hdwallet-keepkey/src/binance.ts
@@ -10,6 +10,7 @@ import {
   BinanceTransferMsg,
 } from "@keepkey/device-protocol/lib/messages-binance_pb";
 import { MessageType } from "@keepkey/device-protocol/lib/messages_pb";
+import BigNumber from "bignumber.js";
 
 export function binanceGetAccountPaths(msg: Core.BinanceGetAccountPaths): Array<Core.BinanceAccountPath> {
   return [
@@ -45,8 +46,13 @@ export async function binanceSignTx(
     );
     if (resp.message_type === Core.Events.FAILURE) throw resp;
 
+    const outputAmount = new BigNumber(message.outputs[0].coins[0].amount);
+    const inputAmount = new BigNumber(message.inputs[0].coins[0].amount);
+    if (!outputAmount.isInteger()) throw new Error("Output amount must be an integer");
+    if (!inputAmount.isInteger()) throw new Error("Input amount must be an integer");
+
     let coinOut = new BinanceTransferMsg.BinanceCoin();
-    coinOut.setAmount(message.outputs[0].coins[0].amount.toString());
+    coinOut.setAmount(outputAmount.toString());
     coinOut.setDenom(message.outputs[0].coins[0].denom);
 
     let outputs = new BinanceTransferMsg.BinanceInputOutput();
@@ -54,7 +60,7 @@ export async function binanceSignTx(
     outputs.setCoinsList([coinOut]);
 
     let coinIn = new BinanceTransferMsg.BinanceCoin();
-    coinIn.setAmount(message.inputs[0].coins[0].amount.toString());
+    coinIn.setAmount(inputAmount.toString());
     coinIn.setDenom(message.inputs[0].coins[0].denom);
 
     let inputs = new BinanceTransferMsg.BinanceInputOutput();
--- a/packages/hdwallet-native/package.json
+++ b/packages/hdwallet-native/package.json
@@ -18,6 +18,7 @@
     "@shapeshiftoss/hdwallet-core": "1.12.1",
     "@terra-money/terra.js": "^1.6.0",
     "bchaddrjs": "^0.4.9",
+    "bignumber.js": "^9.0.1",
     "bip32": "^2.0.5",
     "bip39": "^3.0.2",
     "bitcoinjs-lib": "^5.1.10",
--- a/packages/hdwallet-native/src/binance.ts
+++ b/packages/hdwallet-native/src/binance.ts
@@ -6,6 +6,7 @@ import { NativeHDWalletBase } from "./native";
 import { getNetwork } from "./networks";
 import { toWords, encode } from "bech32";
 import CryptoJS, { RIPEMD160, SHA256 } from "crypto-js";
+import BigNumber from "bignumber.js";
 import util from "./util";
 
 export function MixinNativeBinanceWalletInfo<TBase extends core.Constructor>(Base: TBase) {
@@ -88,11 +89,13 @@ export function MixinNativeBinanceWallet<TBase extends core.Constructor<NativeHD
           throw Error("Invalid permissions to sign for address")
         }
         const addressTo = msg.tx.msgs[0].outputs[0].address;
-        const amount = (msg.tx.msgs[0].inputs[0].coins[0].amount).toString();
+        // The Binance SDK takes amounts as decimal strings.
+        const amount = new BigNumber(msg.tx.msgs[0].inputs[0].coins[0].amount);
+        if (!amount.isInteger()) throw new Error("amount must be an integer");
         const asset = "BNB";
         const memo = msg.tx.memo;
 
-        const result: any = await client.transfer(addressFrom, addressTo, amount, asset, memo, null);
+        const result: any = await client.transfer(addressFrom, addressTo, amount.shiftedBy(-8).toString(), asset, memo, null);
         const pub_key = result.signatures[0].pub_key.toString("base64");
         const signature = Buffer.from(result.signatures[0].signature, "base64").toString("base64");
 
--- a/integration/src/binance/tx02.mainnet.signed.json
+++ b/integration/src/binance/tx02.mainnet.signed.json
@@ -40,6 +40,6 @@
   "signatures": {
     "pub_key": "eb5ae9872102ddf08b1f51a4132c63c73455cac6569404c78c9e3b5bffd1f64e07a9a2f109b4",
     "signature": "bMqcWsPQaNKstYqP8+poAHEgkh6uzH9KLDHP3pt1PUwVjK5XGaA6z2Uk2lfKjZFjSQPG2sGD94nwiDhfd4THjg==",
-    "kksignature": "3Bk42J0FCCt7VSm3OVdn918IL3Z6bKqDzUxKy/eyd1JmJF+Qbd5Vv65YkqVWcK5xEVrwjFMC69I0WFwobySu0w=="
+    "kksignature": "JpznJ+cEUW/L8VUc3zGO1fLQYnzSG1qP11LsrN8ax+hhCag7Wr1MZC3nzdNDD2eHlFJQJXb7LdpF0l2rih29jw=="
   }
 }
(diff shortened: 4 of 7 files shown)


packages/hdwallet-native/package.json
```diff
@@ -18,6 +18,7 @@
     "@shapeshiftoss/hdwallet-core": "1.12.1",
     "@terra-money/terra.js": "^1.6.0",
     "bchaddrjs": "^0.4.9",
+    "big.js": "^6.0.3",
```
majorhayes, line 21: we use `bignumber.js` everywhere else. I don't know the difference but I'd prefer to be consistent.

packages/hdwallet-native/src/binance.ts
```diff
@@ -88,7 +89,8 @@ export function MixinNativeBinanceWallet<TBase extends core.Constructor<NativeHD
           throw Error("Invalid permissions to sign for address")
         }
         const addressTo = msg.tx.msgs[0].outputs[0].address;
-        const amount = (msg.tx.msgs[0].inputs[0].coins[0].amount).toString();
+        // The Binance SDK takes amounts as decimal strings.
+        const amount = Big(msg.tx.msgs[0].inputs[0].coins[0].amount).div(Big(10).pow(8)).toString();
```
majorhayes, line 93: has the data structure already been validated at this point? i.e., if it's malformed and `inputs` is empty, this will hard crash.

Also, I would recommend doing an `isNaN()` check prior to `toString()` and raise an error if it's NaN.

packages/hdwallet-keepkey/src/binance.ts
```diff
@@ -45,16 +45,22 @@ export async function binanceSignTx(
     );
     if (resp.message_type === Core.Events.FAILURE) throw resp;
 
+    const intRegex = /^(0|([1-9][0-9]*))$/;
```
majorhayes, line 48: If we're importing `bignumber.js` we might as well use their already tested function to determine in the value is a whole number

https://mikemcl.github.io/bignumber.js/#isInt

packages/hdwallet-keepkey/src/binance.ts
```diff
@@ -45,16 +45,22 @@ export async function binanceSignTx(
...
+    const outputAmount = message.outputs[0].coins[0].amount.toString();
+    if (!intRegex.test(outputAmount)) throw new Error("output amount should be an integer");
+    const inputAmount = message.inputs[0].coins[0].amount.toString();
+    if (!intRegex.test(inputAmount)) throw new Error("input amount should be an integer");
```
majorhayes, line 52:
```suggestion
    const intRegex = /^(0|([1-9][0-9]*))$/;
    const outputAmount = BigNumber(message.outputs[0].coins[0].amount) // maybe need optional chaining or other validation
    const inputAmount = BigNumber(message.inputs[0].coins[0].amount)
    // BigNumber.prototype.isInteger returns false for NaN values
    if (!outputAmount.isInteger()) throw new Error("Output amount must be an integer")
    if (!inputAmount.isInteger()) throw new Error("Input amount must be an integer")
```

packages/hdwallet-native/src/binance.ts
natven: good point majorhayes. if it's a flaw, it's an existing one. when we're talking about wallet functionality, amounts, addresses etc it might be desirable to just hard crash if something is off vs alternatives like signing the wrong amount.

Does Big have a `Big.pow10(n)`?

packages/hdwallet-keepkey/src/binance.ts
```diff
@@ -45,16 +45,22 @@ export async function binanceSignTx(
...
+
     let coinOut = new BinanceTransferMsg.BinanceCoin();
-    coinOut.setAmount(message.outputs[0].coins[0].amount.toString());
+    coinOut.setAmount(outputAmount);
```
majorhayes, line 55: If using my previous suggestion
```suggestion
    coinOut.setAmount(outputAmount.toString());
```

packages/hdwallet-keepkey/src/binance.ts
```diff
@@ -45,16 +45,22 @@ export async function binanceSignTx(
...
     coinOut.setDenom(message.outputs[0].coins[0].denom);
 
     let outputs = new BinanceTransferMsg.BinanceInputOutput();
     outputs.setAddress(message.outputs[0].address);
     outputs.setCoinsList([coinOut]);
 
     let coinIn = new BinanceTransferMsg.BinanceCoin();
-    coinIn.setAmount(message.inputs[0].coins[0].amount.toString());
+    coinIn.setAmount(inputAmount);
```
majorhayes, line 63: If using my previous suggestion
```suggestion
    coinIn.setAmount(inputAmount.toString());
```

packages/hdwallet-native/package.json
reidrankin: I used `big.js` because it's what the BNB SDK uses to do its math, so it would be guaranteed to be available in any project using the SDK. No problem switching though.

packages/hdwallet-native/src/binance.ts
reidrankin: IMHO hard crash good, malformed data structures bad. none of the structures are really validated consistently anywhere, honestly; that seems to be more of a systemic problem. FWIW on the NaN front, `big.js` is a subset of `bignumber.js`, and doesn't do NaN or infinities at all.

packages/hdwallet-keepkey/src/binance.ts
reidrankin: if we make it available in hdwallet-keepkey i agree. was trying to optimize to import in the smallest scope possible; didn't want to import a whole lib just to get that function. OTOH we could for example `bignumber.js` in the workspace deps, and use it here for no additional cost. Thoughts?

packages/hdwallet-keepkey/src/binance.ts
reidrankin: pedantic nitpick: optional chaining is the opposite of validation. validation explicitly restricts the set of accepted inputs to a function; optional chaining explicitly expands the set of accepted inputs.

packages/hdwallet-keepkey/src/binance.ts
majorhayes: `bignumber.js` is a very small package but I understand the desire to not include another dependency. In our case, since we already use `bignumber.js` yarn/npm will hoist it resulting in no need for an additional copy of the dependency to be installed.

In this case I'm leaning more toward using the package as it has been well tested/hardened and less likely to have a gotcha.

However, if you create some tests for this code that covers all kinds of invalid inputs and you can validate that it works as expected, then I would be okay with that (though I would recommend refactoring out into a shared helper function).

reidrankin: i'd prefer the workspace-root approach myself, just didn't have enough background knowledge to feel comfortable doing it that way. will refactor and rebase.

reidrankin: sigh -- we have a lot of dev deps at the workspace level, but no non-dev ones. i'm guessing that's for a good reason, b/c i don't think the eventually-published npm packages will include workspace-level deps. I'll just add `bignumber.js` to both packages instead.

packages/hdwallet-keepkey/src/binance.ts
majorhayes: In this case it works because `undefined` would return `NaN` and thus be invalided by the `isInteger` check. I would consider that valid value validation in this particular case.

It may be worth looking at doing proper object structure validation at the beginning of the operation in the future.

majorhayes (approved):
I don't want to hold this up. I've said my piece.

reidrankin:
sorry if I gave you the wrong impression! i agree about `bignumber.js`; just slow to push the (correct) change since i apparently angered git at some point here

packages/hdwallet-native/src/binance.ts
```diff
@@ -88,7 +89,8 @@ export function MixinNativeBinanceWallet<TBase extends core.Constructor<NativeHD
           throw Error("Invalid permissions to sign for address")
         }
         const addressTo = msg.tx.msgs[0].outputs[0].address;
-        const amount = (msg.tx.msgs[0].inputs[0].coins[0].amount).toString();
+        // The Binance SDK takes amounts as decimal strings.
+        const amount = new BigNumber(msg.tx.msgs[0].inputs[0].coins[0].amount).shiftedBy(-8).toString();
```
majorhayes, line 93: do we need an `isNaN` check here ?

reidrankin: yes we do. I misread the docs on `shiftedBy()` and thought it would throw.

reidrankin: added an `isInteger()` check before the shift, just for good measure.

reidrankin: that will also make sure anything feeding in a decimal -- like it would have had to do to work around this bug -- will fail fast.

reidrankin:
ok, pretty sure this is ready to go now. reminder -- whoever hits merge needs to run `lerna publish` and add a bump to the minor version afterwards

majorhayes (approved):

PR #271, opened 2021-04-02
Improved Tests Phase 2: Improved Tests

reidrankin:
Following on from the groundwork laid in #265, #270, #251, #252, #253, and #250 (:facepalm:), this is a test-only PR which adds full test coverage<sup>1</sup> to `hdwallet-native`<sup>2</sup>.

<sup>1</sup>: Offer of full coverage only applies to code committed by v1.10.0 or thereabouts. Offer may not yet include Thorchain, Terra, Kava, Secret, or any other strangely-named coins.
<sup>2</sup>: Claims of "Full Coverage" exclude the last commit, which removes all the tests which don't work because they expose parts of the code that is broken.

### DIFF
--- a/packages/hdwallet-native/__mocks__/@shapeshiftoss/hdwallet-core.ts
+++ b/packages/hdwallet-native/__mocks__/@shapeshiftoss/hdwallet-core.ts
@@ -1,0 +1,7 @@
+// This ensures that the module's exports are all writable, so that jest.spyOn() will work.
+const mock = {};
+const actual = jest.requireActual("@shapeshiftoss/hdwallet-core");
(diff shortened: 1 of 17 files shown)


reidrankin:
the `mswMock` and `untouchableMock` stuff could stand to be factored out to separate packages at some point.

reidrankin:
The idea of the last commit is to revert all the tests that don't pass because of bugs. That ensures that this PR touches only testing-related code, and doesn't require a version bump. The next PR in the series will contain the fixes for the uncovered issues, as well as a commit reverting the reversion.

This helps separate the concerns of testing behavior and fixing behavior, and is also objectively hilarious.

BitHighlander (approved):
:shipit:

PR #272, opened 2021-04-05
remove terra/kava/secret sdk's

BitHighlander:
Remove sdk's that break react native support


reidrankin (changes requested):
These should be something like `throw new Error("not implemented")` instead of `return null`.

BitHighlander:
> These should be something like `throw new Error("not implemented")` instead of `return null`.

patched

reidrankin (approved):

mvpratt (approved):

PR #282, opened 2021-05-05
Isolated crypto for hdwallet-native

reidrankin:
This adds machinery to isolate cryptographic operations from the libraries and SDKs used by hdwallet-native. Doing so requires some minor tweaks to these libraries, which are applied by `patch-package`. The isolation code includes:

- a core module which contains type-checking machinery and cryptographic interfaces which make available the full range of possible operations with each type of isolated cryptographic primitive
- adapters that, well, adapt these interfaces to interoperate with specific third-party libraries
- a "dummy" isolation engine, which stores secrets in ES2019 private fields and performs operations using `tiny-secp256k1` and `hmacSHA512` from `bip32/src/crypto`. (better name pending? it's supposed to mean not-actually-isolated, rather than doesn't-actually-work.)

Future work will involve the creation of additional engines which proxy through to native platform code where applicable, as well as a WASM (or maybe just Web Worker) engine to ensure secrets are held in a separate memory space even on web.


BitHighlander (approved):
:shipit:

PR #287, opened 2021-05-18
Tenderment tx building

BitHighlander:
Terra/Kava/Secret support. with a migration to tendermint-tx-builder npm. 

https://www.npmjs.com/package/tendermint-tx-builder


packages/hdwallet-native/src/terra.ts
```diff
@@ -70,7 +71,11 @@ export function MixinNativeTerraWallet<TBase extends core.Constructor<NativeHDWa
```
mvpratt, line 71:
```
terraSupportsNativeShapeShift(): boolean {      return false;    }
```
should this be true instead?  (line 19 above)

packages/hdwallet-native/src/kava.test.ts
```diff
@@ -63,6 +63,14 @@ describe("NativeKavaWallet", () => {
       chain_id: "foobar",
       account_number: "foo",
       sequence: "bar",
-    })).rejects.toThrowError("Not Supported");
+    })
+    await expect(signed.signatures.length).toBe(1);
```
kaladinlight, line 67: You don't need to await expect on the already resolved promise

.circleci/config.yml
```diff
@@ -53,6 +53,9 @@ orbs:
           - run:
               name: Ensure repo is clean from uncommitted changes, likely yarn.lock is out of date
               command: git diff --quiet || exit 1
+#          - run:
```
kaladinlight, line 56: Not working? Might just want to add in a separate PR

kaladinlight (approved):

mcchadwick (approved):

PR #291, opened 2021-06-01
The Mega Patch

reidrankin:
- various housekeeping things
- native wallet crypto isolation improvments
  - fix some bugs
  - eliminate some patches
- safer Buffer.from() and ArrayBuffer conversions
- `hdwallet-ledger` no longer reinvents as many wheels
- mocked network responses prevent tests from failing when the wind is blowing at the wrong speed
- `xxxGetAddress()` no longer requires the address as a parameter for `showDisplay === true`
- more detailed types
- strict mode turned on
- major import/dependency cleanup
  - also various dependency version bumps
- `hdwallet-keepkey-electron` is now a thing


integration/package.json
```diff
@@ -10,21 +10,23 @@
...
+    "@shapeshiftoss/hdwallet-core": "^1.15.0",
+    "@shapeshiftoss/hdwallet-keepkey": "^1.15.0",
+    "@shapeshiftoss/hdwallet-keepkey-nodewebusb": "^1.15.0",
+    "@shapeshiftoss/hdwallet-keepkey-tcp": "^1.15.0",
+    "@shapeshiftoss/hdwallet-trezor": "^1.15.0",
+    "@types/debug": "^4.1.5",
+    "@types/jest": "^26.0.23",
+    "debug": "^4.2.0",
     "fast-json-stable-stringify": "^2.1.0",
+    "jest": "^26.6.3",
```
kaladinlight, line 22: Seems  like some weird dependency inclusions here. Can we make use the of the global dev dependencies for the likes of jest/lerna/typescript etc, and then pull out any non required dependencies into devDependencies?

integration/src/bitcoin/bitcoin.ts
```diff
@@ -217,66 +203,67 @@ export function bitcoinTests(get: () => { wallet: HDWallet; info: HDWalletInfo }
...
           },
         ];
-        let outputs = [
+        const outputs: core.BTCSignTxOutput[] = [
           {
             address: "1MJ2tj2ThBE62zXbBYA5ZaN3fdve5CPAz1",
-            addressType: BTCOutputAddressType.Spend,
-            scriptType: BTCOutputScriptType.PayToAddress,
+            addressType: core.BTCOutputAddressType.Spend,
+            // scriptType: core.BTCOutputScriptType.PayToAddress,
```
kaladinlight, line 258: Reason for not including script type?

.circleci/config.yml
```diff
@@ -5,10 +5,10 @@ orbs:
     executors:
       node:
         docker:
-          - image: circleci/node:12
+          - image: circleci/node:14
```
majorhayes, line 8: Why? All the projects that import this module use node 12, so I feel like this has the opportunity to create some incompatibilities.

integration/jest.config.js
```diff
@@ -9,9 +9,11 @@ module.exports = {
   moduleNameMapper: {
     "^@shapeshiftoss/(.*)": "<rootDir>/../../packages/$1/src",
   },
-  testPathIgnorePatterns: ["dist"],
   globals: {
-    Uint8Array: Uint8Array,
-    ArrayBuffer: ArrayBuffer,
+    'ts-jest': {
+      diagnostics: {
+        ignoreCodes: [7016],
```
reidrankin, line 15: This is "could not find a declaration file for module" and is caused because of how ts-jest chooses files to include in the compilation process. It will suppress errors during `yarn test`, but that's OK because they'll still occur during `yarn build` if they're valid.

integration/package.json
reidrankin: The integration tests need all this stuff to actually run, so some things you'd expect to be in `devDependencies` are actually in `dependencies`. (Not that anyone's probably going to install the integration test package by itself, but if they did, they'd need `jest` to run the package, not just dev on it.)

integration/src/bitcoin/bitcoin.ts
reidrankin: There's two ways to specify an address: either as a literal address in the `address` field, or calculated from the wallet's seed via `addressNList` and `scriptType`. When `address` is present, `scriptType` is not used at all, because the format of the address string implies the script type. The new strict type stuff uses a helper (`GuardedUnion<T>`) to cause an error to result if you try to specify fields that are defined on a different union member than you're actually using, so removing the `scriptType` is required.

integration/src/bitcoin/bitcoin.ts
```diff
@@ -294,74 +281,74 @@ export function bitcoinTests(get: () => { wallet: HDWallet; info: HDWalletInfo }
...
           },
           {
-            address: "1FH6ehAd5ZFXCM1cLGzHxK1s4dGdq1JusM",
-            addressType: BTCOutputAddressType.Change,
-            scriptType: BTCOutputScriptType.PayToAddress,
+            addressNList: core.bip32ToAddressNList("m/44'/0'/0'/0/0"),
+            addressType: core.BTCOutputAddressType.Change,
+            scriptType: core.BTCOutputScriptType.PayToAddress,
             amount: String(9000),
-            isChange: false,
```
reidrankin, line 358: Here's some evidence of the type confusion we've had going -- passing integration tests with `addressType: BTCOutputAddressType.Change` and `isChange: false`. (WTF?)

integration/src/bitcoin/testnet.ts
```diff
@@ -44,94 +31,100 @@ export function testnetTests(get: () => { wallet: HDWallet; info: HDWalletInfo }
...
 
     test("btcSignTx() - p2wpkh", async () => {
-      if (!wallet || isPortis(wallet)) return;
-      if (isLedger(wallet)) return; // FIXME: Expected failure
+      if (!wallet || portis.isPortis(wallet)) return;
+      if (ledger.isLedger(wallet)) return; // FIXME: Expected failure
       if (!wallet.btcSupportsCoin("Testnet")) return;
-      let inputs = [
+      const tx: core.BitcoinTx = {
+        version: core.untouchable("tx.version not provided by test"),
```
reidrankin, line 82: `core.untouchable` provides magical `as any`-like powers, but without sacrificing safety.

integration/src/bitcoin/testnet.ts
```diff
@@ -44,94 +31,100 @@ export function testnetTests(get: () => { wallet: HDWallet; info: HDWalletInfo }
...
-            vout: [
-              {
-                scriptPubKey: {
-                  hex: "0014b31dc2a236505a6cb9201fa0411ca38a254a7bf1",
-                },
-              },
-            ],
-          },
+          tx,
+          hex: undefined as string,
```
reidrankin, line 102: We can't fill in `hex` from `tx` through the power of manual calculation because `tx` is missing a bunch of fields. (We should probably replace this test with something from the actual blockchain so we don't have to pull these shenanigans.)

packages/hdwallet-core/src/bitcoin.ts
```diff
@@ -37,63 +47,179 @@ export interface BitcoinOutput {
...
+  scriptType: BTCInputScriptType;
   hex: string;
-  type?: number; // Dash
-  extraPayloadSize?: number; // Dash
-  extraPayload?: string; // Dash
-}
+};
+
+export type BTCSignTxInput = BTCSignTxInputNative & BTCSignTxInputKK & BTCSignTxInputTrezor & BTCSignTxInputLedger;
+export type BTCSignTxInputUnguarded = BTCSignTxInputNativeUnguarded & BTCSignTxInputKKUnguarded & BTCSignTxInputTrezor & BTCSignTxInputLedger;
```
reidrankin, line 132: This dichotomy is dumb. We should fix all btcSignTx implementations to use a single representation of the input tx -- probably `hex` -- and eliminate the confusion.

Of course there are probably edge-cases in clients which are passing in inconsistent `tx` and `hex` values -- so normalizing how we actually process the data, rather than just how we type it, may change client behavior and will need a deft hand.

packages/hdwallet-core/src/bitcoin.ts
```diff
@@ -37,63 +47,179 @@ export interface BitcoinOutput {
...
+>;
 
 export interface BTCSignTx {
   coin: string;
   inputs: Array<BTCSignTxInput>;
   outputs: Array<BTCSignTxOutput>;
   version?: number;
   locktime?: number;
-  opReturnData?: string;
+  opReturnData?: string; // TODO: dump this in favor of BTCSignTxOutputMemo above
```
reidrankin, line 214: dumping this in is semi-urgently needed before it gets implemented -- and used -- in more places.

packages/hdwallet-core/src/bitcoin.ts
```diff
@@ -37,63 +47,179 @@ export interface BitcoinOutput {
...
   vaultAddress?: string;
 }
 
+export type BTCSignTxKK = Omit<BTCSignTx, "inputs"> & { inputs: Array<BTCSignTxInputKK> };
+export type BTCSignTxNative = Omit<BTCSignTx, "inputs"> & { inputs: Array<BTCSignTxInputNative> };
+export type BTCSignTxTrezor = Omit<BTCSignTx, "inputs"> & { inputs: Array<BTCSignTxInputTrezor> };
+export type BTCSignTxLedger = Omit<BTCSignTx, "inputs"> & { inputs: Array<BTCSignTxInputLedger> };
```
reidrankin, line 221: as discussed above, normalize BTCSignTxInput and we can dump these special types.

packages/hdwallet-core/src/utils.ts
```diff
@@ -29,33 +29,33 @@ export function toHexString(arr: Uint8Array): string {
...
-    if (!match) {
-      throw new Error("invalid hexidecimal string");
-    }
+  let match = value.match(/^(0x)?[0-9a-fA-F]*$/);
 
-    if (match[1] !== "0x") {
-      throw new Error("hex string must have 0x prefix");
-    }
+  if (!match) {
+    throw new Error("invalid hexidecimal string");
```
reidrankin, line 41: oh look typo
```suggestion
    throw new Error("invalid hexadecimal string");
```

patches/@types+tiny-secp256k1+1.0.0.patch
```diff
@@ -0,0 +1,20 @@
...
+ export function sign(message: Buffer, privateKey: Buffer): Buffer;
+ 
++/**
++ * Returns normalized signatures, each of (r, s) values are guaranteed to less than order / 2. Uses RFC6979.
++ * Adds e as Added Entropy to the deterministic k generation.
++ * @param message should be 256-bit
++ * @param privateKey isPrivate(privateKey) should be true
++ * @param entropy should be 256-bit
++ */
++ export function signWithEntropy(message: Buffer, privateKey: Buffer, entropy: Buffer): Buffer;
```
reidrankin, line 16: This function is present in tiny-secp256k1 but is missing in the DefinitelyTyped repo for some reason.

BitHighlander:
Vercel broke :( need that for hosting sandbox

reidrankin:
Unfortunately, all I get from the Vercel "details" link is a 404, so I don't know what's wrong

integration/jest.config.js
majorhayes: Include this explanation as a comment in the file please

reidrankin:
(there's at least a couple changes to make, so I'm reverting to draft pending completion)

integration/jest.config.js
reidrankin:
```suggestion
        // TS(7016) "could not find a declaration file for module" occurs when ts-jest is confused about which *.d.ts to pass
        // as input to TypeScript's compilation API. When this happens, noImplicitAny will complain. Disabling this error will
        // suppress it when run under ts-jest; this is OK, because it will still be reported during the normal build process.
        ignoreCodes: [7016],
```

integration/src/bitcoin/bitcoin.ts
reidrankin: in practice, this strangeness is an attempt to reconcile `hdwallet-keepkey`'s version of `validateVoutOrdering` -- which requires `BTCOutputAddressType.Change` in the second position -- with the usage of a literal change address in the test. This shouldn't work, because the whole point of `BTCOutputAddressType.Change` is to be used with an `addressNList` and `scriptType` for the wallet to calculate the address. (This is required so that your KK knows you own the address of the change output; otherwise it wouldn't be safe for it to subtract the change amount from the amount it has you confirm the spending of.)      _just so happens_ that validate

.circleci/config.yml
reidrankin: We should be able to handle any specific backwards-compatibility requirements handled by ensuring the `tsconfig.json` compilation options are set appropriately.

More to the point, though, node 12 is two years old by now. It's already out of active support, and will be completelly EOL next year. Something's gotta be the first thing updated -- and hdwallet has better test coverage than some of its prominent dependents do. (Besides, Node 14 has WASI, and that will probably come in handy if/when there's a WASM engine for the isolated crypto stuff.)

mvpratt (approved):

BitHighlander (approved):
:shipit:

PR #292, opened 2021-06-08
patch ethSignTx

BitHighlander:
This patch's the way hdwallet assembles the final tx to allow correct input of "0x0" as a nonce for the 0th tx on a new ethereum account.


packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -71,7 +71,9 @@ export async function ethSignTx(transport: KeepKeyTransport, msg: ETHSignTx): Pr
   return transport.lockDuring(async () => {
     const est: EthereumSignTx = new EthereumSignTx();
     est.setAddressNList(msg.addressNList);
-    est.setNonce(arrayify(msg.nonce));
+    let nonce = arrayify(msg.nonce);
+    while (nonce.length > 0 && nonce[0] == 0x00) nonce = nonce.slice(1);
+    est.setNonce(nonce);
```
mcchadwick, line 76: You could also do something like:

```suggestion
    const idx = arrayify(msg.nonce).findIndex((x) => x != 0x00);
    est.setNonce(idx > -1 ? msg.nonce?.slice(idx) : [0x00]);
```
I'm not sure which is stylistically preferable.

mcchadwick (approved):

PR #295, opened 2021-06-10
avoid prompt ReferenceError if not in browser 

reidrankin:
Even if it's not used, importing `kdwallet-keepkey-webusb` in a non-browser context (read: node) will cause a ReferenceError. Some things do that, so this stops them from breaking immediately.


BitHighlander (approved):

reidrankin:
Closing in favor of #299.

PR #297, opened 2021-06-10
Work around bad Buffer.concat polyfills

reidrankin:
Webpack 4's `Buffer.concat()` polyfill requires `Buffer[]` instead of `Uint8Array[]`. This is a kludgy compatibility hack until everything gets bumped to Webpack 5.


reidrankin:
Closing in favor of #299.

PR #298, opened 2021-06-10
Fix addresses in BNB sandbox example Tx

reidrankin:
BNB transaction signing would fail in the sandbox with "Invalid permissions to sign for address" because the listed input address does not match the default mnemonic; now it doesn't.


reidrankin:
Closing in favor of #299.

PR #299, opened 2021-06-10
some compatibility fixes

reidrankin:
This clears up a few minor issues:
- Importing `hdwallet-keepkey-webusb` in a non-browser context would fail because `window` doesn't exist. Obviously this package doesn't work in those kinds of contexts, but it no longer throws merely on import.
- Webpack 4 depends on `node-libs-browser` which is very out-of-date. In fact, the `Buffer` polyfill it pulls in doesn't support `Buffer.concat()` on `Uint8Array[]`, only on `Buffer[]`. This is wrong, and the fix is to use Webpack 5, but until that's possible in all downstream packages there's now a kludgy compatibility hack.
- Replaced a `while` loop with a `findIndex` call
- BNB sandbox benchmark Tx now doesn't always fail with `Invalid permissions to sign for address`


reidrankin:
Closing in favor of #303.

PR #302, opened 2021-06-11
vercel test 3

reidrankin:


reidrankin:
The Vercel config has now been updated so that it doesn't hide the logs anymore.

PR #316, opened 2021-08-11
Add support for EIP-1559 fees to hdwallet-native

mcchadwick:


majorhayes (approved):

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -102,6 +108,12 @@ export interface ETHWalletInfo extends HDWalletInfo {
    */
   ethSupportsNativeShapeShift(): boolean;
 
+  /**
+   * 
+   * Does the device support transactions with EIP-1559 fee parameters?
+   */
+  ethSupportsEIP1559(): boolean;
```
reidrankin, line 115: I'd like to see this done up as a proper type guard... I'll do up an example

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -102,6 +108,12 @@ export interface ETHWalletInfo extends HDWalletInfo {
    */
   ethSupportsNativeShapeShift(): boolean;
 
+  /**
+   *
```
reidrankin, line 112:
```suggestion
   *
```

mvpratt (approved):

packages/hdwallet-core/src/ethereum.ts
reidrankin: (After discussion with @majorhayes and @mcchadwick, this will be handled as a fast-follow)

PR #317, opened 2021-08-13
Type guards for EIP-1559

mrnerdhair:


packages/hdwallet-core/src/ethereum.ts
```diff
@@ -1,12 +1,6 @@
 import { addressNListToBIP32, slip44ByCoin } from "./utils";
 import { ExchangeType, BIP32Path, HDWallet, HDWalletInfo, PathDescription } from "./wallet";
```
mvpratt, line 3: I think this enum should stay in core.  the tx types aren't unique to the Native wallet.  just happens that's the only place we are using them so far.  

will confirm shortly whether they are needed for keepkey

packages/hdwallet-native/src/ethereum.ts
```diff
@@ -5,6 +5,12 @@ import _ from "lodash";
 import { NativeHDWalletBase } from "./native";
 import * as Isolation from "./crypto/isolation";
 
+export enum EthersTransactionType {
```
reidrankin, line 8:
```suggestion
enum EthersTransactionType {
```

packages/hdwallet-core/src/ethereum.ts
reidrankin: The issue is that we don't need these values at all, except to interact with `ethers`, and so we shouldn't export the type if we can help it. (I forgot to remove the `export` after moving the enum, but that was the point).

Note that `ethers`'s use of `0` to indicate a legacy transaction instead of an EIP-2718 TransactionType actually breaks the spec -- `0` is a valid, non-legacy TransactionType, it just hasn't been used by anything. The spec is clear:
> If it starts with a value in the range [0, 0x7f] then it is a new transaction type, if it starts with a value in the range [0xc0, 0xfe] then it is a legacy transaction type.

There's no reason for users to ever need to specify a TransactionType. EIP-2930 is used to include an `accessList` field, and it would be unambiguous as to whether an `ETHSignTx` which contained that field should be signed according to that EIP; likewise, EIP-1559 signing is unambiguously required when there's a `maxFeePerGas` field.

reidrankin: (I just discovered in #349 that specifying `ETH_TX_TYPE_LEGACY` actively breaks KK.)

0xean:
Given the staleness of the PR, I am going to close this for now and we can re-open at a future date if need be.

PR #325, opened 2021-09-14
Test eip1559 native

markrypto:


packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -984,6 +985,9 @@ export class KeepKeyHDWallet implements core.HDWallet, core.BTCWallet, core.ETHW
     this._supportsEos = semver.gte(fwVersion, "v6.4.0");
     // this._supportsThorchain = Semver.get(fwVersion, "v7.0.0");
 
+    // EIP-1559 isn't supported until v7.2.1
+    this._supportsEip1559 = semver.gte(fwVersion, "v7.2.1");
+
```
mvpratt, line 990: 👍 
didn't realize we already had this pattern

packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -984,6 +985,9 @@ export class KeepKeyHDWallet implements core.HDWallet, core.BTCWallet, core.ETHW
     this._supportsEos = semver.gte(fwVersion, "v6.4.0");
     // this._supportsThorchain = Semver.get(fwVersion, "v7.0.0");
 
+    // EIP-1559 isn't supported until v7.2.1
+    this._supportsEip1559 = semver.gte(fwVersion, "v7.2.1");
```
mvpratt, line 989:
```suggestion
    this._supportsEthEip1559 = semver.gte(fwVersion, "v7.2.1");
```

packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -611,6 +611,7 @@ export class KeepKeyHDWallet implements core.HDWallet, core.BTCWallet, core.ETHW
   _supportsRipple = true;
   _supportsBinance = true;
   _supportsEos = true;
+  _supportsEip1559 = false;
```
mvpratt, line 614:
```suggestion
  _supportsEthEip1559 = false;
```

packages/hdwallet-keepkey/src/keepkey.ts
mvpratt: I think it would be good to add the chain name prefix.  So later if we have something like `_supportsCosmosXyz` or `_supportsEthAbcFeature`

mvpratt:
need to bump the packages to an alpha version, publish that to nom and test before merging this @markrypto .  I can help w/ this

packages/hdwallet-keepkey/src/keepkey.ts
cjthompson: we should probably do this for other things like `_supportsThorChain` in another PR

cjthompson (approved):

PR #328, opened 2021-09-20
make eip1559 query async to work for keepkey

markrypto:


packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -21,8 +21,8 @@ export function ethSupportsNativeShapeShift(): boolean {
   return true;
 }
 
-export function ethSupportsEIP1559(): boolean {
-  return this.ethSupportsEIP1559();
+export async function ethSupportsEIP1559(): Promise<boolean> {
```
majorhayes, line 24: This breaks the way all of the "Supports" functions work. This is why we did firmware checks in axiom instead of hdwallet.

I don't like the idea of having some "Supports" functions be async and some be sync.

packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -21,8 +21,8 @@ export function ethSupportsNativeShapeShift(): boolean {
...
+  return await this.ethSupportsEIP1559();
```
cjthompson, line 25: This breaks the way all of the "Supports" functions work. This is why we did firmware checks in axiom instead of hdwallet.

I don't like the idea of having some "Supports" functions be async and some be sync.

markrypto: "Supports" functions are all hard coded booleans which doesn't seem right either. From my point of view, directly querying the wallet is the most direct, efficient and understandable way to obtain feature support, especially for the hardware wallets.

mvpratt: the drawback of checking in Axiom is all other clients would need to do the same

mvpratt: > I don't like the idea of having some "Supports" functions be async and some be sync.

agreed.  can we make them all async?

packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -985,9 +984,6 @@ export class KeepKeyHDWallet implements core.HDWallet, core.BTCWallet, core.ETHW
     this._supportsEos = semver.gte(fwVersion, "v6.4.0");
     // this._supportsThorchain = Semver.get(fwVersion, "v7.0.0");
 
-    // EIP-1559 isn't supported until v7.2.1
-    this._supportsEthEip1559 = semver.gte(fwVersion, "v7.2.1");
-
```
mvpratt, line 990: I forget, did we get to the root cause of why this wasn't working?

mvpratt (approved):

PR #329, opened 2021-09-20
bech32 support native

BitHighlander:


packages/hdwallet-native/src/bitcoin.test.ts
```diff
@@ -133,7 +133,7 @@ const OP_RETURN_BENCHMARK_TX = benchmarkTx(
   { opReturnData: "foobar" }
 );
 
-describe("NativeBTCWalletInfo", () => {
+describe.only("NativeBTCWalletInfo", () => {
```
cjthompson, line 136:
```suggestion
describe("NativeBTCWalletInfo", () => {
```

yarn.lock
```diff
@@ -2876,6 +2876,15 @@
     web-encoding "^1.1.0"
     wif "^2.0.6"
 
+"@shapeshiftoss/hdwallet-core@1.15.5-alpha.1":
```
cjthompson, line 2879: Why is hdwallet importing a copy of itself in the yarn.lock file?

PR #339, opened 2021-09-30
bugfix: fix ledger pairing/syncing issues

millsmcilroy:


packages/hdwallet-ledger-webusb/src/adapter.ts
```diff
@@ -60,15 +60,13 @@ export class WebUSBLedgerAdapter {
     }, APP_NAVIGATION_DELAY);
   }
 
-  public get(device: USBDevice): core.HDWallet | null {
-    if (!device.serialNumber) throw new Error("missing serial number");
-    return this.keyring.get(device.serialNumber);
+  public get(device: USBDevice): core.HDWallet {
+    return this.keyring.get((device as any).serialNumber);
```
millsmcilroy, line 64: Ledgers do not have serial numbers

packages/hdwallet-ledger/src/bitcoin.ts
```diff
@@ -72,13 +72,13 @@ export async function btcGetPublicKeys(
...
     const parentPublicKey = compressPublicKey(Buffer.from(parentPublicKeyHex, "hex"));
-    const parentFingerprint = new DataView(bitcoin.crypto.ripemd160(bitcoin.crypto.sha256(parentPublicKey))).getUint32(0);
+    const parentFingerprint = new DataView(bitcoin.crypto.ripemd160(bitcoin.crypto.sha256(parentPublicKey)).buffer).getUint32(0);
 
     const res2 = await transport.call("Btc", "getWalletPublicKey", bip32path, opts);
     handleError(res2, transport, "Unable to obtain public key from device.");
 
     const {
-      payload: { publicKeyHex, chainCodeHex },
+      payload: { publicKey: publicKeyHex, chainCode: chainCodeHex },
```
millsmcilroy, line 81: This was a bug, `publicKeyHex` and `chainCode` were coming back `undefined`

packages/hdwallet-ledger-webusb/src/adapter.ts
cjthompson: If that's true, won't this crash or return undefined?

I'd rather just return a static value.

packages/hdwallet-ledger/src/bitcoin.ts
```diff
@@ -72,13 +72,13 @@ export async function btcGetPublicKeys(
       payload: { publicKey: parentPublicKeyHex },
     } = res1;
     const parentPublicKey = compressPublicKey(Buffer.from(parentPublicKeyHex, "hex"));
-    const parentFingerprint = new DataView(bitcoin.crypto.ripemd160(bitcoin.crypto.sha256(parentPublicKey))).getUint32(0);
+    const parentFingerprint = new DataView(bitcoin.crypto.ripemd160(bitcoin.crypto.sha256(parentPublicKey)).buffer).getUint32(0);
```
cjthompson, line 75: Buffer can return unexpected data because of shared heap but since you're getting the first value, it might be okay? @reidrankin ?

packages/hdwallet-ledger/src/utils.ts
```diff
@@ -71,15 +71,15 @@ export const compressPublicKey = (publicKey: Uint8Array) => {
 };
 
 export const createXpub = (depth: number, parentFp: number, childNum: number, chainCode: Uint8Array, publicKey: Uint8Array, network: number) => {
-  const header = new Uint8Array(4 + 1 + 4 + 4);
+  const header = new Uint8Array(4 + 1 + 4 + 4).buffer;
```
cjthompson, line 74: @reidrankin ?

cjthompson (approved):

packages/hdwallet-ledger-webusb/src/adapter.ts
millsmcilroy: Yeah....good catch. I was waiting for it to break but it didn't. On the axiom side we hardcode all ledgers to have deviceId aka serialNumber as "0001" Maybe we should hardcode it here too? @kaladinlight Do you have any ideas?

packages/hdwallet-ledger/src/bitcoin.ts
millsmcilroy: The `DataView` constructor was not happy with its first argument being a plain `Buffer`. It wants an `ArrayBufferLike`

packages/hdwallet-ledger-webusb/src/adapter.ts
kaladinlight: The `0001` id is hardcoded on the ledger device itself. Whether it is the serialNumber or deviceID or both, I don't recall exactly. All I remember is there was no good unique identifiers at the time, so we just used the same one for everyone and said we weren't going to support multiple ledgers connected at once. Would have to query the device on the `window.navigator` to verify. My ledger doesn't even connect to ledger live anymore....

millsmcilroy: Kevin is right. During pairing it was throwing the "missing serial number" error which broke the whole pairing flow, but I confirmed by removing that throw then after the pair step we are able to access device.serialNumber which is "0001"

Issue #306, opened 2021-07-13
Proposal: Usage of `ledger-thorchain-js`


veado:
[`ledger-thorchain-js`](https://github.com/thorchain/ledger-thorchain-js) will be the "official" JS Ledger library for THORChain.  Since `hdwallet` is already using [`@ledgerhq/hw-app-btc`](https://www.npmjs.com/package/@ledgerhq/hw-app-btc) and [`@ledgerhq/hw-app-eth`](https://www.npmjs.com/package/@ledgerhq/hw-app-eth), `ledger-thorchain-js` might be a good  option as well.

reidrankin:
Thanks for opening this issue -- I know we talked in Discord but this definitely helps track things better. I expect we'll be looking to add support for the THORChain Ledger app soon, and this may be the best way.

PR #352, opened 2021-10-16
draft: ledger thorchain support

mrnerdhair:
Based on #351.

As an example of how type information helps speed development, here's a thing I hacked together in like an hour that makes a start on #306. No testing on real devices yet, and not sure how to handle the fact that core really wants `getPublicKeys` to return an xpub, but the point is that I imported a completely new ledger app package with its own set of completely new quirks and wired the thing up with autocomplete and tooltips in no time flat.


0xean:
closing as stale. @mrnerdhair  - please re-open as necessary.

PR #354, opened 2021-10-18
Refactor various architectural things

mrnerdhair:
Based on #350 and #351.

With the advent of web v2, we have an opportunity to make a breaking API change and release an hdwallet 2.0. This is a collection of a bunch of stuff that would be nice to do, at least some of which changes the external API enough that it'll probably break an existing client.

This isn't done yet, but it's ready for some comments.


0xean:
closing as stale. @mrnerdhair  - please re-open as necessary.

PR #358, opened 2021-10-19
Various minor improvements

reidrankin:
These are the parts of #354 that are non-breaking. Each commit is single-purpose and should be reviewable in isolation.

### DIFF
--- a/packages/hdwallet-core/src/transport.ts
+++ b/packages/hdwallet-core/src/transport.ts
@@ -15,7 +15,7 @@ export abstract class Transport extends eventemitter2.EventEmitter2 {
   /**
    * Must emit outgoing message events and communicate with underlying interface
    */
-  public abstract call(...args: any[]): Promise<any>;
+  public abstract call(...args: any): Promise<unknown>;
 
   /**
    * Optional method to bootstrap connection to device
--- a/packages/hdwallet-core/src/utils.ts
+++ b/packages/hdwallet-core/src/utils.ts
@@ -208,3 +208,36 @@ export function compatibleBufferConcat(list: Uint8Array[]): Buffer {
   if (!checkBufferConcat()) return Buffer.concat(list);
   return Buffer.concat(list.map(x => Buffer.isBuffer(x) ? x : Buffer.from(x)));
 }
+
+
+/**
+ * Type guard for things that might have (string-keyed) properties. Useful to make
+ * TypeScript happy when you want to check if an object of unknown type has a particular
+ * property set.
+ * @example
+ * try {
+ *   foo();
+ * } catch (e: unknown) {
+ *   // Not allowed because there's no index signature for `unknown`:
+ *   // if (e.bar === "baz") return "foobar";
+ *   if (isIndexable(e) && e.bar === "baz") return "foobar";
+ *   throw e;
+ * }
+ * @example
+ * isIndexable({}) === true
+ * @example
+ * isIndexable(() => {}) === true
+ * @example
+ * isIndexable(Object.create(null)) === true
+ * @example
+ * isIndexable(String("foo")) === true
+ * @example
+ * isIndexable(null) === false
+ * @example
+ * isIndexable(3.14) === false
+ * @example
+ * isIndexable("foo") === false
+ */
+export function isIndexable(x: unknown): x is Record<string | number | symbol, unknown> {
+  return x !== null && ["object", "function"].includes(typeof x)
+}
--- a/packages/hdwallet-keepkey-webusb/src/adapter.ts
+++ b/packages/hdwallet-keepkey-webusb/src/adapter.ts
@@ -32,7 +32,7 @@ export const AdapterDelegate = {
       if (out.serialNumber === undefined) throw new Error("expected serial number");
       return out as Device;
     } catch (e) {
-      throw new core.WebUSBCouldNotPair("KeepKey", e.message);
+      throw new core.WebUSBCouldNotPair("KeepKey", String(core.isIndexable(e) ? e.message : e));
     }
   },
   async getTransportDelegate(device: Device) {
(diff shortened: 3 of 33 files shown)


packages/hdwallet-core/src/utils.ts
```diff
@@ -208,3 +208,7 @@ export function compatibleBufferConcat(list: Uint8Array[]): Buffer {
   if (!checkBufferConcat()) return Buffer.concat(list);
   return Buffer.concat(list.map(x => Buffer.isBuffer(x) ? x : Buffer.from(x)));
 }
+
+export function isIndexable(x: unknown): x is Record<string, unknown> {
+  return x !== null && ["object", "function"].includes(typeof x)
```
cjthompson, line 213: save yourself a few CPU cycles and check undefined up-front?
```suggestion
  return x != null && ["object", "function"].includes(typeof x)
```

packages/hdwallet-core/src/utils.ts
```diff
@@ -208,3 +208,7 @@ export function compatibleBufferConcat(list: Uint8Array[]): Buffer {
   if (!checkBufferConcat()) return Buffer.concat(list);
   return Buffer.concat(list.map(x => Buffer.isBuffer(x) ? x : Buffer.from(x)));
 }
+
+export function isIndexable(x: unknown): x is Record<string, unknown> {
```
cjthompson, line 212: 1. I do not like the name. It took some time to guess what it was for
2. No JSDoc block with an explanation of what the function does. For a utility function with a specific purpose, a JSDoc comment would be incredibly useful.

packages/hdwallet-keepkey-webusb/src/adapter.ts
```diff
@@ -32,7 +32,7 @@ export const AdapterDelegate = {
       if (out.serialNumber === undefined) throw new Error("expected serial number");
       return out as Device;
     } catch (e) {
-      throw new core.WebUSBCouldNotPair("KeepKey", e.message);
+      throw new core.WebUSBCouldNotPair("KeepKey", String(core.isIndexable(e) ? e.message : e));
```
cjthompson, line 35: nit: I'd prefer you just import the function and use it as `isIndexable` instead of `core.isIndexable`.

packages/hdwallet-trezor/src/trezor.ts
```diff
@@ -258,28 +248,11 @@ export class TrezorHDWalletInfo implements core.HDWalletInfo, core.BTCWalletInfo
...
-  readonly _supportsFio = false;
-  readonly _supportsThorchainInfo = false;
-  readonly _supportsThorchain = false;
-  readonly _supportsSecretInfo = false;
-  readonly _supportsSecret = false;
-  readonly _supportsKava = false;
   readonly _supportsKavaInfo = true;
-  readonly _supportsTerra = false;
   readonly _supportsTerraInfo = true;
+  readonly _isTrezor = true;
```
cjthompson, line 255: Are you sure this isn't going to break anything? I would not be surprised in the least if there's code that directly references these instead of using the helper functions.

cjthompson:
My **primary** concern here is the removal of the `_supportsX` properties. There's a non-zero chance that code somewhere references those instead of using the "proper" `supportsX` functions. To me the risk/reward leans heavily in favor of not removing them.

Best leave that for a major revision.

packages/hdwallet-core/src/utils.ts
reidrankin: I'd argue that comparing with to `null` or `undefined` with `!=` is dangerous in general and the benefits of avoiding having to think about why this instance is OK outweigh any performance impact.

packages/hdwallet-core/src/transport.ts
```diff
@@ -15,7 +15,7 @@ export abstract class Transport extends eventemitter2.EventEmitter2 {
   /**
    * Must emit outgoing message events and communicate with underlying interface
    */
-  public abstract call(...args: any[]): Promise<any>;
+  public abstract call(...args: any): Promise<unknown>;
```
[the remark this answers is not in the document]
reidrankin: There is a subtle difference -- `any[]` is an array _of a homogenous type_. For example, you can assign a `[string, number]` to `any[]`, but only by narrowing it to `(string|number)[]`, not all the way down to `[string, number]` or `[string, number , ...any[]]` like you might expect. The standard library uses `any` instead of `any[]` for stuff like `Parameters<T>` for this reason.

packages/hdwallet-core/src/utils.ts
reidrankin: 1) I agree, but I can't think of a better one
2) Good point, agree, will update

packages/hdwallet-keepkey-webusb/src/adapter.ts
reidrankin: We should move the function out of `hdwallet-core` to a utility package at some point, but I od feel `import { * as core, isIndexable } from "@shapeshiftoss/hdwallet-core";` would be the more awkward syntax.

packages/hdwallet-keepkey/src/binance.ts
```diff
@@ -50,10 +50,11 @@ export async function binanceSignTx(
     let resp = await transport.call(
       Messages.MessageType.MESSAGETYPE_BINANCESIGNTX,
       signTx,
-      core.LONG_TIMEOUT,
-      /*omitLock=*/ true
+      {
+        msgTimeout: core.LONG_TIMEOUT,
+        omitLock: true
+      }
```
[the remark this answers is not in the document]
reidrankin: `transport.call` is internal to the `hdwallet-keepkey-*` family of packages; below, I add yet-another optional parameter to it, and it made sense to collect them in container where they can all be explicitly named.

packages/hdwallet-trezor/src/trezor.ts
reidrankin: Code that references these will still get `undefined` -- things will only break if something's doing `=== false` explicitly.

packages/hdwallet-core/src/utils.ts
reidrankin: Now with 100% more JSDoc!

PR #366, opened 2021-11-04
native: improve isolated crypto interfaces

reidrankin:
This updates the isolated crypto interfaces so that all methods (and fields) are async. The previous definition supported an async implementation, but would have required a complicated wrapper to ensure that a couple of operations "looked" synchronous. Eliminating that requirement comes primarily from having touched enough of the code to realize that `buildInput()` can be made async safely.

This also updates all the native wallet so that it can be initialized with an `Isolation.Core.Bip32.Node` instead of an `Isolation.Core.Bip32.Mnemonic`, which permits simpler isolation engine implementations, and renames the `Dummy` engine to `Default` to better reflect its role and robustness.


packages/hdwallet-native/src/bitcoin.ts
```diff
@@ -215,10 +215,10 @@ export function MixinNativeBTCWallet<TBase extends core.Constructor<NativeHDWall
       return true;
     }
 
-    buildInput(coin: core.Coin, input: core.BTCSignTxInputNative): InputData | null {
-      return this.needsMnemonic(!!this.#seed, () => {
+    async buildInput(coin: core.Coin, input: core.BTCSignTxInputNative): Promise<InputData | null> {
```
reidrankin, line 218: This is the core reason things weren't like this before -- because `buildInput()` was synchronous and technically part of the public interface. That had trickle-down effects that prevented `util.getKeyPair()` from being async and from there propagated out.

Now that I'm more comfortable that I have a full picture of the the overall library and its use cases, I'm confident that this method should be marked `protected` if it weren't for the fact that we're unable to do so because the mixin strategy we use is build on exporting class expressions, which doesn't support `protected` or `private`. Therefore, I don't think this counts as a modification to the public API, so we can change it without breaking things.

packages/hdwallet-native/src/crypto/isolation/adapters/binance.ts
```diff
@@ -3,15 +3,15 @@ import { BIP32, Digest, SecP256K1 } from "../core"
 
 type SigningDelegate = Parameters<BncClient["setSigningDelegate"]>[0];
 
-export function signingDelegate(keyPair: BIP32.Node): SigningDelegate {
+export default {
+  async create(keyPair: BIP32.Node): Promise<SigningDelegate> {
```
reidrankin, line 7: This helps normalize the way that isolation adapters are used -- via `await Isolation.Adapters.Binance.create()` just like any other isolation adapter.

PR #369, opened 2021-11-12
Fix MetaMask Connect Flow

pastaghost:


packages/hdwallet-metamask/src/adapter.ts
```diff
@@ -55,6 +44,17 @@ export class MetaMaskAdapter {
   }
 
   private async pairMetaMaskDevice(): Promise<core.HDWallet> {
```
cjthompson, line 46: I don't see a need for this separate private function. Just make this the `public async pairDevice()` function.

cjthompson (approved):

PR #388, opened 2022-01-13
feat: add XDeFi Adapter

loxator:
Addition of XDeFi Adapter to work with the XDeFi Wallet. Currently, only supporting the Ethereum chain, and will be used to add more chains further.


packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+    switch (msg.coin) {
+      case "Ethereum":
+        return eth.describeETHPath(msg.path);
+      default:
+        throw new Error("Unsupported path");
+    }
+  }
+
+  public ethNextAccountPath(msg: core.ETHAccountPath): core.ETHAccountPath | undefined {
+    // TODO: What do we do here?
```
reidrankin, line 309: This method isn't implemented in anything else, and it's really only in hdwallet for legacy reasons; account paths are (will be?) taken care of by chain-adapters now. (We need to do an API scrub now that we've dropped support for the old beta.shapeshift.com stuff.)

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  public hasOnDeviceRecovery(): boolean {
+    return true;
+  }
+
+  public hasNativeShapeShift(srcCoin: core.Coin, dstCoin: core.Coin): boolean {
+    // It doesn't... yet?
+    return false;
+  }
```
reidrankin, line 289: This is all legacy BS and will be burned with fire at some point.

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+    return chainId === 1;
+  }
+
+  public async ethSupportsSecureTransfer(): Promise<boolean> {
+    return false;
+  }
+
+  public ethSupportsNativeShapeShift(): boolean {
+    return false;
+  }
```
reidrankin, line 323: (fyi, legacy BS again)

packages/hdwallet-xdefi/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "main": "dist/index.js",
+  "source": "src/index.ts",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "tsc --build",
+    "clean": "rm -rf dist tsconfig.tsbuildinfo",
+    "prepublishOnly": "yarn clean && yarn build"
+  },
+  "dependencies": {
+    "@portis/web3": "3.0.10",
```
reidrankin, line 17: betting some of these deps need cleanup

packages/hdwallet-xdefi/src/ethereum.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  let index = path[2] & 0x7fffffff;
+  return {
+    verbose: `Ethereum Account #${index}`,
+    accountIdx: index,
+    wholeAccount: true,
+    coin: "Ethereum",
+    isKnown: true,
+  };
+}
```
reidrankin, line 32: You can get this logic from `core.describeETHPath()`

packages/hdwallet-xdefi/src/ethereum.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export async function ethVerifyMessage(msg: core.ETHVerifyMessage, ethereum: any): Promise<boolean | null> {
+  console.error("Method ethVerifyMessage unsupported for XDeFi wallet!");
+  return null;
+}
```
reidrankin, line 37: This can be polyfilled in if we can get the public key out of xdefi

packages/hdwallet-xdefi/src/adapter.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  public async initialize(): Promise<number> {
+    return Object.keys(this.keyring.wallets).length;
+  }
+
+  public async pairDevice(): Promise<core.HDWallet> {
+    const provider: any = (window as any).xfi && (window as any).xfi.ethereum;
+    if (!provider) {
+      window.open("https://xdefi.io", "_blank");
+    }
```
reidrankin, line 26: It's not appropriate to fire window.open here; better to throw and let the application handle it. My instinct is also to use `globalThis` instead of `window` but idk.

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+    return !this.provider.xdefi.isUnlocked();
+  }
+
+  public getVendor(): string {
+    return "XDeFi";
+  }
+
+  public getModel(): Promise<string> {
+    return Promise.resolve("XDeFi");
+  }
```
reidrankin, line 74:
```suggestion
  public async getModel(): Promise<string> {
    return "XDeFi";
  }
```

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+
+  public getLabel(): Promise<string> {
+    return Promise.resolve("XDeFi");
+  }
```
reidrankin, line 78:
```suggestion
  public async getLabel(): Promise<string> {
    return "XDeFi";
  }
```

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+
+  public async initialize(): Promise<any> {
+    try {
+      this.provider = (window as any).xfi && (window as any).xfi.ethereum;
+    } catch (e) {
+      console.error(e);
+    }
+
+    return Promise.resolve();
+  }
```
reidrankin, line 88: returning `any` kills kittens
```suggestion
  public async initialize(): Promise<void> {
    try {
      this.provider = (window as any).xfi && (window as any).xfi.ethereum;
    } catch (e) {
      console.error(e);
    }
  }
```

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+
+  public recover(msg: core.RecoverDevice): Promise<void> {
+    // no concept of recover in XDeFi
+    return Promise.resolve();
+  }
+
+  public loadDevice(msg: core.LoadDevice): Promise<void> {
+    // TODO: Does XDeFi allow this to be done programatically?
```
reidrankin, line 166: (probably not, and that's fine)

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+  readonly _supportsFioInfo = false;
+  readonly _supportsFio = false;
+  readonly _supportsThorchainInfo = false;
+  readonly _supportsThorchain = false;
+  readonly _supportsSecretInfo = false;
+  readonly _supportsSecret = false;
+  readonly _supportsKava = false;
+  readonly _supportsKavaInfo = false;
+  readonly _supportsTerra = false;
+  readonly _supportsTerraInfo = false;
```
reidrankin, line 49: - nothing good will come of breaking the naming convention
> _alles muss in ordnung kommen_
- all the `= false` can die
```suggestion
  readonly _supportsETH = true;
  readonly _isXDeFi = true;
```

reidrankin (changes requested):
First off, thanks for this PR! We appreciate it greatly and look forward to getting it in. I've taken a first-pass at this and left some suggestions; it may be easiest to drop into the Engineering section of our Discord server and workshop this live.

Also, one thing that we will need before merge is a comprehensive set of unit and integration tests, which will probably require cooking up some sort of xdefi mock if they don't have one already.

loxator:
Hi @reidrankin,

Thanks for the awesome review and raising some very valid points. I am resolving them one by one as I go, some are pretty straightforward, but I would like a bit of clarification on a couple of them.

The Adapter which was committed is a copy of your MetaMask adapter. The idea is to start of XDeFi with the EVM based chains, and then move onto other chains.

You have marked some of the methods as `obsolete` but they still exist in the interface that the adapter implements. For example, some of the methods come from the interface `core.HDWallet`, since you mentioned that they are going to be removed at some point, should I leave the methods in for now, and can be refactored later on when the interface changes? Or should I also remove them from both the places in this PR?

Also, it'd be great if there is an example I can look at for implementing the unit/integration tests so I conform to your repository's standards.

Thanks a lot again for taking the time to review and going over the PR, I am still new to the world of wallets and learning as I go along, so I will have a lot of questions going forward :sweat_smile: !

packages/hdwallet-xdefi/src/adapter.ts
loxator: You have a very good point, I will throw an exception here and handle it on the client side. Will also check between `this` vs `window`.

reidrankin:
Sorry for the confusion about the obsolescence of various things; they do still have to be implemented when required by the interfaces, but hopefully knowing our plans to nuke them will give you some useful context.

For test references, check out `integration/src/*.test.ts`, `integration/src/wallets/*.ts`. You'll need to:
- implement any appropriate unit tests for your new package; our best-tested packages are `hdwallet-native` and `hdwallet-keepkey`, so take a look at `packages/hdwallet-{native,keepkey}/src/*.test.ts` to get an idea of what's needed
- cut-n-paste `integration/src/foobar.ts` to `integration/src/xdefi.ts`
- implement the `WalletSuite` interface from `integration/src/wallets/suite.ts` in `integration/src/wallets/xdefi.ts`; I'd start by copying the portis one, since that's a web3 sort of thing too
-  you'll probably need to mock the parts of xdefi you need to make your code testable; tests must be consistently reproducible, which means they must pass without a network connection

As it turns out, `hdwallet-metamask` lacks these sorts of tests; I remember it was implemented under a certain amount of time pressure, which is probably why we dropped the ball. However, I have a feeling that making the tests needed for `hdwallet-xdefi` and `hdwallet-metamask` will be highly similar, and I have another feeling that our engineering workstream will be more than willing to throw some dosh in your general direction for eliminating this particular oversight. (@0xdef1cafe back me up on this?)

packages/hdwallet-xdefi/src/adapter.ts
```diff
@@ -0,0 +1,42 @@
...
+  public static useKeyring(keyring: core.Keyring) {
+    return new XDeFiAdapter(keyring);
+  }
+
+  public async initialize(): Promise<number> {
+    return Object.keys(this.keyring.wallets).length;
+  }
+
+  public async pairDevice(): Promise<core.HDWallet> {
+    const provider: any = (window as any).xfi && (window as any).xfi.ethereum;
```
reidrankin, line 23:
```suggestion
    const provider: any = (globalThis as any).xfi?.ethereum;
```

packages/hdwallet-xdefi/src/adapter.ts
```diff
@@ -0,0 +1,42 @@
...
+      window.open("https://xdefi.io", "_blank");
+    }
+    try {
+      await provider.request({ method: "eth_requestAccounts" });
+    } catch (error) {
+      console.error("Could not get XDeFi accounts. ");
+      throw error;
+    }
+    const wallet = new XDeFiHDWallet();
+    await wallet.initialize();
```
reidrankin, line 34: Just noticed that `provider` doesn't actually get passed into the instance here!

packages/hdwallet-xdefi/src/ethereum.ts
reidrankin: Actually, now that I think about it, we don't need the public key -- we can just recover it from the message signature, calculated the address from the key, and compare it with what it's supposed to be

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+  readonly _supportsETHInfo = true;
+  readonly _supportsCosmosInfo = false;
+  readonly _supportsBinanceInfo = false;
+  readonly _supportsRippleInfo = false;
+  readonly _supportsEosInfo = false;
+  readonly _supportsFioInfo = false;
+  readonly _supportsThorchainInfo = false;
+  readonly _supportsSecretInfo = false;
+  readonly _supportsKavaInfo = false;
+  readonly _supportsTerraInfo = false;
```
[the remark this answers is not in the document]
reidrankin:
```suggestion
  readonly _supportsETHInfo = true;
```

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,340 @@
...
+    return [];
+  }
+
+  public async isInitialized(): Promise<boolean> {
+    return true;
+  }
+
+  public disconnect(): Promise<void> {
+    return Promise.resolve();
+  }
```
reidrankin, line 185:
```suggestion
  public async disconnect(): Promise<void> { }
```

packages/hdwallet-xdefi/src/adapter.ts
loxator: The `provider` is not passed as that's how it the method signature is defined in `code.HDWallet`. Should the provider be passed into the instance instead?

```
  /**
   * Initialize a device session.
   */
  initialize(): Promise<any>;
  ```

File: packages/hdwallet-xdefi/src/xdefi.ts (hunk 16 above)
Comment (loxator) [reply packages/hdwallet-xdefi/src/xdefi.ts:185]: This needs a promise to be resolved as that's how the signature is in the interface `core.HDWallet`, otherwise Typescript complains.

**Comment by loxator** [comment]
@reidrankin 

I have tried to add unit and integration tests according to your instructions.

I was able to get 100% coverage for most of the unit test files, except `xdefi.ts`. I have left this one for now as a lot of methods are just Promises being resolved due to the implemented interface for `core.HDWallet`.

I have tried to set up the integration tests as you mentioned, but have a bit of difficulty understanding  on how to change some tests that`suite.ts` runs. For example, our wallet doesn't support verification of signed messages, yet there is a test in `ethereumTests` that `suites.ts` runs on `xdefi.ts` in the integration tests, how do I overwrite the necessary tests?

Lastly, thanks a lot for your help!

**Comment by loxator** [comment]
@reidrankin,

I have added the necessary tests and also implemented the missing functions for `ethVerifyMessage` and `ethSignTransaction`. 

Please let me know if anything else is required from my end? Once this adapter is merged, then I can proceed with the Bitcoin Adapter. If changes are required, please let me know soon as the XDeFi team would like this to be included as soon as possible. 

Thanks a lot for all your help!

**Comment by loxator** [comment]
@reidrankin @pastaghost 

Just bumping this PR up as we would like to add the integration for EVM chains soon. Thank you :pray: !

File: packages/hdwallet-xdefi/src/xdefi.ts (hunk 16 above)
Comment (cjthompson) [reply packages/hdwallet-xdefi/src/xdefi.ts:158]: `async` guarantees a `Promise` is returned so I would think this should be fine.

Comment (loxator) [reply packages/hdwallet-xdefi/src/xdefi.ts:158]: That is true, but we aren't returning anything in the suggestion right? Typescript complains that a value must be returned 
`A function whose declared type is neither 'void' nor 'any' must return a value.ts(2355)`

Comment (loxator) [reply packages/hdwallet-xdefi/src/xdefi.ts:158]: Please ignore the previous comment as I missed out on the `async` bit.

**Comment by loxator** [comment]
@cjthompson I have updated the PR according to your comments.

**Comment by cjthompson** [comment]
@mrnerdhair - the build is failing because it can't find the `@shapeshiftoss/hdwallet-xdefi` package, which obviously isn't published. I would think that it would be better for the build to use the local code in `packages` vs downloading published packages. Can we look at fixing module resolution to resolve locally vs. pulling from npm?

**Review by pastaghost — APPROVED** [review]

**Comment by mrnerdhair** [comment]
@loxator I think I've fixed the build issue here. I've PRed the fix into this branch: https://github.com/XDeFi-tech/hdwallet/pull/1

File: packages/hdwallet-xdefi/src/adapter.ts (hunk 14 above)
Comment (mrnerdhair) [reply packages/hdwallet-xdefi/src/adapter.ts:34]: Ah, gotcha. We can still extend that interface, we just have to do it a bit wonkily -- see the matching comment on the wallet implementation for details.
```suggestion
    const wallet = new XDeFiHDWallet();
    await wallet.initialize(provider);
```

packages/hdwallet-xdefi/src/index.test.ts
```diff
@@ -0,0 +1,11 @@
+import * as library from "./";
+
+describe("Exports all expected classes", () => {
+  it("should export EncryptedWallet", () => {
```
mrnerdhair, line 4:
```suggestion
  it("should export XDeFiAdapter", () => {
```

packages/hdwallet-xdefi/src/index.test.ts
```diff
@@ -0,0 +1,11 @@
...
+    expect(library.XDeFiAdapter.name).toBe("XDeFiAdapter");
+  });
+
+  it("should export WebCryptoEngine", () => {
```
mrnerdhair, line 8:
```suggestion
  it("should export XDeFiHDWallet", () => {
```

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -0,0 +1,288 @@
...
+    return "XDeFi";
+  }
+
+  public async initialize(): Promise<any> {
+    try {
+      this.provider = (window as any).xfi && (window as any).xfi.ethereum;
+    } catch (e) {
+      console.error(e);
+    }
+  }
```
mrnerdhair, line 59: The first overload provides a static hint that the method will throw without options given. The implementation signature complies with the core interface but isn't visible itself from other code, avoiding any ambiguity about `undefined` being a supported value.

```suggestion
  public initialize(): never;
  public initialize(provider: unknown): Promise<any>;
  public async intialize(provider?: unknown): Promise<any> {
    if (!provider) throw new Error("provider is required");
    this.provider = provider;
  }
```

mrnerdhair (approved):
Excellent work here. I've PRed the build fix as XDeFi-tech#1; once that is integrated and these comments are resolved, LGTM.

mrnerdhair:
Migrated to #405.

PR #389, opened 2022-01-14
fix: gas limit parameter for MetaMask

Neurone:
Fixed Metamask parameter for `gasLimit`.

Metamask and other web3 libraries (i.e. `geth console`, `web3.js`, etc.) use `transaction.gas` property for transaction's gas limit, instead of the canonical `transaction.gasLimit` as mentioned in the [yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf). This implies Metamask does not read any value for gas limit from the JSON object, and so it chooses one on its own. This behavior can lead to inconsistencies between clients and the wallet, for example when forecasting the transaction's cost (see https://github.com/shapeshift/web/issues/526).

### DIFF
--- a/packages/hdwallet-metamask/src/ethereum.ts
+++ b/packages/hdwallet-metamask/src/ethereum.ts
@@ -63,6 +63,10 @@ export async function ethSendTx(msg: core.ETHSignTx, ethereum: any, from: string
       data: msg.data,
       chainId: msg.chainId,
       nonce: msg.nonce,
+      // MetaMask, like other Web3 libraries, derives its transaction schema from Ethereum's official JSON-RPC API specification
+      // (https://github.com/ethereum/execution-apis/blob/d63d2a02bcd2a8cef54ae2fc5bbff8b4fac944eb/src/schemas/transaction.json).
+      // That schema defines the use of the label `gas` to set the transaction's gas limit and not `gasLimit` as used in other
+      // libraries and as stated in the official yellow paper (https://ethereum.github.io/yellowpaper/paper.pdf).
       gas: msg.gasLimit,
     };
 
--- a/packages/hdwallet-metamask/src/ethereum.ts
+++ b/packages/hdwallet-metamask/src/ethereum.ts
@@ -63,7 +63,7 @@ export async function ethSendTx(msg: core.ETHSignTx, ethereum: any, from: string
       data: msg.data,
       chainId: msg.chainId,
       nonce: msg.nonce,
-      gasLimit: msg.gasLimit,
+      gas: msg.gasLimit,
     };
 
     const utx = msg.maxFeePerGas


Neurone:
Please note I run the test suite before any modification, and after the fix, and the result is the following:

```shell
Test Suites: 3 failed, 24 passed, 27 total
Tests:       20 failed, 302 passed, 322 total
Snapshots:   24 passed, 24 total
Time:        432.043 s
```

reidrankin:
Just fyi, there shouldn't be any failing tests -- something might be busted with your setup?

Neurone:
I just did `yarn, yarn build, yarn test`. I think the integration tests are failing.

Documentation says:

> The integration tests have been set up to run either against a physical KeepKey
> with debug firmware on it, or in CI pointed at a dockerized version of the
> emulator. 

I don't have a KeepKey (sorry ^^) and I don't know how startup an emulator. CI via Github have all checks passed, so I suppose this is the reason.

Neurone:
Just noticed the `docker:run:emulator` in the package.json ^^ Anyway, all problem seems related to `argonBenchmark` and the `require("crypto").webcrypto` object that seems undefined, I'll try to scout more as soon as I have more time to dedicate:

```shell
  ● argonBenchmark › argonBenchmark › should run correctly

    TypeError: Cannot read property 'subtle' of undefined

      13 |     const realCrypto = require("crypto").webcrypto as Crypto;
      14 |     setCrypto({
    > 15 |       subtle: realCrypto.subtle,
         |                          ^
      16 |       getRandomValues: await deterministicGetRandomValues(realCrypto),
      17 |     });
```

Neurone:
The problem was with my `Nodejs v14.18.3  (Latest LTS: Fermium)`. I switched to `v16.13.2 (Latest LTS: Gallium)` and now all tests work just fine!

reidrankin:
That makes sense; hdwallet supports being used under older node versions but does require node 16 for tests. As far as the KK emulator goes, you can
```
git clone https://github.com/keepkey/keepkey-firmware
cd keepkey-firmware
git submodule update --init --recursive
cd scripts/emulator
docker-compose up kkemu
```
to build/run it; I think the `docker:run:emulator` version is a bit old at this point.

cjthompson (approved):

packages/hdwallet-metamask/src/ethereum.ts
```diff
@@ -63,7 +63,7 @@ export async function ethSendTx(msg: core.ETHSignTx, ethereum: any, from: string
       data: msg.data,
       chainId: msg.chainId,
       nonce: msg.nonce,
-      gasLimit: msg.gasLimit,
+      gas: msg.gasLimit,
```
cjthompson, line 66: Can you add a comment here with a link to a source that documents this behavior for future reference?

0xean (approved):

packages/hdwallet-metamask/src/ethereum.ts
Neurone: Done

PR #391, opened 2022-01-17
feat: lazy-load various things

mrnerdhair:
In general, loading the third-party SDKs we need is quite a time-consuming operation (>500ms in dev mode!), and it's not strictly necessary until they need to be actually used. Using dynamic `import()` to pull them in allows this process to be deferred until a convenient moment so that the UI isn't blocked.

Note that the commit to `hdwallet-native-vault` retains a static dependency on `hdwallet-native` because it contains `Revocable`, which somewhat defeats the point. However, this will soon (#400) be refactored out of `hdwallet-native`, which will remove this dependency and realize the performance potential.

### DIFF
--- a/integration/src/wallets/portis.ts
+++ b/integration/src/wallets/portis.ts
@@ -38,19 +38,19 @@ export async function createWallet(): Promise<core.HDWallet> {
 
   // mock web3.eth
   // this feels bad man, would be better to test against a debug verision of Portis should it ever exist
-  wallet.web3 = {
+  wallet.web3 = Promise.resolve({
     eth: {
       accounts: {
-        recover: () => Promise.resolve("0x3f2329C9ADFbcCd9A84f52c906E936A42dA18CB8"),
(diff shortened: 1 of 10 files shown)


integration/src/wallets/portis.ts
```diff
@@ -38,19 +38,19 @@ export async function createWallet(): Promise<core.HDWallet> {
 
   // mock web3.eth
   // this feels bad man, would be better to test against a debug verision of Portis should it ever exist
-  wallet.web3 = {
+  wallet.web3 = Promise.resolve({
     eth: {
       accounts: {
-        recover: () => Promise.resolve("0x3f2329C9ADFbcCd9A84f52c906E936A42dA18CB8"),
+        recover: () => "0x3f2329C9ADFbcCd9A84f52c906E936A42dA18CB8",
```
mrnerdhair, line 44:
```suggestion
        recover: async () => "0x3f2329C9ADFbcCd9A84f52c906E936A42dA18CB8",
```

cjthompson (approved):

PR #397, opened 2022-01-17
vault: a few minor refactors

mrnerdhair:
These changes create clean extension points for later upgrade to an asynchronous map interface.

PR #398, opened 2022-01-17
vault test improvements

mrnerdhair:
Based on #396 and #397.

This lays the groundwork for testing multiple implementations of the `IVault` interface, and adds a `MockVault` implementation as a reference. While it is intended that only a single implementation will ever be shipped, we will soon need the ability to test it both "raw" and wrapped with a remote proxy of some kind (i.e. in a web worker via comlink).


cjthompson (approved):

packages/hdwallet-native-vault/src/test/mockVault.skip.ts
```diff
@@ -0,0 +1,127 @@
+import * as core from "@shapeshiftoss/hdwallet-core";
```
mrnerdhair, line 1:
```suggestion
/* This is intended to be used to test the tests. It should not be shipped or *shudder* actually used. */
import * as core from "@shapeshiftoss/hdwallet-core";
```

cjthompson (approved):

PR #399, opened 2022-01-17
even-more-async isolation interfaces

mrnerdhair:
The last few properties on the crypto isolation interfaces are now async, enabling them to be used as-is with comlink.

### DIFF
--- a/packages/hdwallet-native-vault/src/index.test.ts
+++ b/packages/hdwallet-native-vault/src/index.test.ts
@@ -98,7 +98,10 @@ function testVaultImpl(name: string, Vault: ISealableVaultFactory<IVault>) {
       const mnemonic = (await vault.get("#mnemonic")) as native.crypto.Isolation.Engines.Default.BIP39.Mnemonic;
       expect(mnemonic).toBeInstanceOf(native.crypto.Isolation.Engines.Default.BIP39.Mnemonic);
       expect(
-        Buffer.from((await (await mnemonic.toSeed()).toMasterKey()).publicKey).toString("hex")
+        await mnemonic.toSeed()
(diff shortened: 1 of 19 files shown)


packages/hdwallet-native-vault/src/index.test.ts
```diff
@@ -98,7 +98,7 @@ function testVaultImpl(name: string, Vault: ISealableVaultFactory<IVault>) {
       const mnemonic = (await vault.get("#mnemonic")) as native.crypto.Isolation.Engines.Default.BIP39.Mnemonic;
       expect(mnemonic).toBeInstanceOf(native.crypto.Isolation.Engines.Default.BIP39.Mnemonic);
       expect(
-        Buffer.from((await (await mnemonic.toSeed()).toMasterKey()).publicKey).toString("hex")
+        Buffer.from(await (await (await mnemonic.toSeed()).toMasterKey()).getPublicKey()).toString("hex")
```
cjthompson, line 101: I'd almost rather see chained .then calls than await await await.

mrnerdhair: something like `await mnemonic.toSeed().then(x => x.toMasterKey()).then(x => x.getPublicKey())`? I could see that

mrnerdhair: resolved as e.g.
 ```
expect(
  await mnemonic.toSeed().then(x => x.toMasterKey()).then(x => x.getPublicKey()).then(x => Buffer.from(x).toString("hex"))
).toMatchInlineSnapshot();
```

cjthompson (approved):

PR #400, opened 2022-01-17
proxy magic

mrnerdhair:
Based on #398.

This hoists the existing `Revocable` machinery into `hdwallet-core` and gives it an overhaul to produce `core.overlay()`, which is a general-purpose proxy doohickey capable not simply of creating magical data rugpull objects like `Revocable` does now but adding, removing, and transforming existing objects (and constructors!) of arbitrary types to add arbitrary functionality.

Notably, this will be used later (#401) to build a set of general-purpose AsyncIterator utilities because it can be used to make objects that are both themselves and promises of themselves. This is the key feature needed to create an object that implements both `Iterator<T>` and `AsyncIterator<T>` at the same time. Being able to convert any iterator from either type into the particular kind needed will make exposing collections over `comlink` connections (i.e. in a web worker) much easier. (At runtime, you can't tell until you actually try to *use* an iterator what type it is... and since comlink happens to turn `Iterator<T>`s into `AsyncIterator<T>`s, we need to be able to write code that can consume both.)

(There's a strong argument to be made that this doesn't actually belong in `hdwallet-core` as much as in some as-yet-indeterminate new library package, but it's not immediately obvious where that should go, and I wanted to get the code out for review.)


cjthompson (dismissed):

0xean:
closing as stale / draft

PR #403, opened 2022-01-22
Add hdwallet-metamask integration tests

### LINKED ISSUE #402: Add hdwallet-metamask integration tests
# Background 
The hdwallet-metamask package does not currently have a set of tests run alongside those for the other supported wallets in the integration test suite. Using the portis and native integration tests as an example, add test coverage for hdwallet-metamask in a metamask.test.ts file and wallets/metamask.ts file under the hdwallet/integration subdirectory. Mock data should be placed in hdwallet/integration/src/wallets/__mocks__.

# Acceptance Criteria
- The wallet should support Ethereum mainnet (add test for `wallet.ethSupportsNetwork()`)
- The wallet should not support BTC (add test for `wallet.btcSupportsNetwork()`)
- The wallet should use the correct ETH BIP-44 paths (add test for `wallet.ethGetAccountPaths()`)
- The wallet should describe ETH paths (add test for `wallet.describeETHPath()`)
- The wallet should return a valid ETH address (add test for `wallet.ethGetAddress()`)
- The wallet should sign a message (add test for `wallet.ethSignMessage()`)

# Note
Since these tests are run outside a browser, the provider property of the wallet will have to be mocked. The following pattern can be extended for any RPC calls made by each of the above functions that call `ethereum.request()`.

```
  // mock wallet provider
  wallet.provider = {
    request: (args: any) => {
      switch (args.method) {
        case "eth_accounts":
          return ["0x3f2329C9ADFbcCd9A84f52c906E936A42dA18CB8"];
          break;
        default:
          return null;
      }
    },
  };
  // end mock
  ```

# Bounty Hunters
- Join our discord [here](https://discord.gg/shapeshift).  
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info

### PR DESCRIPTION
**Pull request by daodev877**
Adds hdwallet-metamask integration tests

- [x] add test for wallet.ethSupportsNetwork()
- [x] add test for wallet.btcSupportsNetwork()
- [x] add test for wallet.ethGetAccountPaths()
- [x] add test for wallet.describeETHPath()
- [x] add test for wallet.ethGetAddress()
- [x] add test for wallet.ethSignMessage()
- [x] mock data should be placed in hdwallet/integration/src/wallets/mocks.

closes #402

### CODE REVIEW

File: integration/src/ethereum/ethereum.ts [hunk 1]
```diff
@@ -119,6 +119,7 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
       "ethSignTx() - ETH",
       async () => {
         if (!wallet) return;
+        if (wallet.getVendor() === 'MetaMask') return;
```
daodev877, line 122: Methods `ethVerifyMessage` and `ethSignTx` are unsupported for MetaMask wallet

cjthompson: I believe there is a function call on the wallet to check is a wallet supports signing like this and you should use that rather than checking vendor name.

cjthompson: See https://github.com/shapeshift/hdwallet/blob/9d1e0b254c29925d8d5e6851097270d17da4146e/packages/hdwallet-metamask/src/metamask.ts#L110

integration/src/wallets/metamask.ts
```diff
@@ -0,0 +1,143 @@
...
+  });
+
+  it("does not support Native ShapeShift", async () => {
+    if (!wallet) return;
+    expect(wallet.ethSupportsNativeShapeShift()).toEqual(false);
+  });
+
+  it("does not support EIP1559", async () => {
+    if (!wallet) return;
+    expect(await wallet.ethSupportsEIP1559()).toEqual(false);
```
0xean, line 48: Is there a reason we aren't supporting EIP1559 with metmask?

0xean (approved):
This looks great, thanks for the contribution! Will let @pastaghost take a look as well.

Left one question re: EIP 1559 which maybe pasta ghost can also answer.

integration/src/ethereum/ethereum.ts
```diff
@@ -119,6 +119,7 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
       "ethSignTx() - ETH",
       async () => {
         if (!wallet) return;
+        if (!wallet.supportsOfflineSigning()) return;
```
pastaghost, line 122: Metamask does not support offline signing, so this statement should force the test to return prematurely. There's a check used in a few places in the lib repository to take conditional action for wallets like metamask that do not support offline signing. Some or part of this check might be useful here instead.

```
  if (wallet.supportsOfflineSigning()) {
    try {
      // Call signTransaction and get signed transaction
    } catch (error) {
      throw new Error(error)
    }

    if (!signedTx) {
      throw new Error(error)
    }

    try {
      // broadcasat previously signed transaction
    } catch (error) {
      throw new Error(error)
    }

    return { txid }
  } else if (wallet.supportsBroadcast() && adapter.signAndBroadcastTransaction) {
    try {
      // Call signAndBroadcastTransaction to sign and broadcast in one step
    } catch (error) {
      throw new Error(error)
    }
    ```

File: integration/src/ethereum/ethereum.ts [hunk 4]
```diff
@@ -148,9 +149,8 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
     test(
       "ethSignTx() - ETH EIP-1559",
       async () => {
-        if (!wallet) {
-          return;
-        }
+        if (!wallet) return;
+        if (!wallet.supportsOfflineSigning()) return;
```
pastaghost, line 153: Same as above.

integration/src/ethereum/ethereum.ts
```diff
@@ -227,6 +227,7 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
       "ethSignTx() - ERC20",
       async () => {
         if (!wallet) return;
+        if (!wallet.supportsOfflineSigning()) return;
```
pastaghost, line 230: Same as above.

integration/src/ethereum/ethereum.ts
```diff
@@ -256,6 +257,7 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
       "ethSignTx() - long contract data",
       async () => {
         if (!wallet) return;
+        if (!wallet.supportsOfflineSigning()) return;
```
pastaghost, line 260: Same as above.

integration/src/ethereum/ethereum.ts
```diff
@@ -350,6 +352,7 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
       "ethVerifyMessage()",
       async () => {
         if (!wallet) return;
+        if (!wallet.supportsOfflineSigning()) return;
```
pastaghost, line 355: Same as above.

integration/src/ethereum/ethereum.ts
daodev877: @pastaghost thanks for the detailed comment! I've updated the tests. 
Could you take another look, please?

integration/src/wallets/metamask.ts
mrnerdhair: (@pastaghost probably knows for sure, but IIRC it was just because it wasn't in the MVP AC)

pastaghost: @0xean At the time the integration was done, EIP-1559-formatted transactions would sign successfully on the MetaMask extension, but would consistently fail when a Ledger/Trezor was used as the signing agent through MetaMask. Trezor has since released a firmware update to fix this, but I don't know about the status of Ledger. If we can get someone to test this on Ledger, we'd be clear to turn the `ethSupportsEIP1559` flag turned on.

pastaghost (approved):

Issue #416, opened 2022-02-11
inject metamask inpage provider so `unsafe-inline` isn't required in the CSP on Firefox


mrnerdhair:
In `hdwallet-metamask`, use `@metamask/providers` (or maybe even the older `metamask-inpage-provider` -- see https://github.com/shapeshift/web/pull/443 for some potentially-adaptable code) to try to instantiate a metamask provider for use in case `detectEthereumProvider()` can't find `window.ethereum`. This is relevant because Firefox applies page CSPs to code injected by extensions, which will block thing metamask's injected provider if we don't include the `unsafe-inline` options.

This may be better served as a change PRed to `@metamask/detect-provider` to resolve https://github.com/MetaMask/detect-provider/issues/31.

(see also https://github.com/shapeshift/web/issues/541)

sudarno08:
Yes

PR #417, opened 2022-02-11
metamask: only detectEthereumProvider() once

mrnerdhair:
This follows the style from the new `hdwallet-xdefi` and makes fixing https://github.com/shapeshift/hdwallet/issues/416 easier.


integration/src/integration.ts
```diff
@@ -21,10 +21,6 @@ import { terraTests } from "./terra";
 import { kavaTests } from "./kava";
 import { WalletSuite } from "./wallets/suite";
 
-import { ethereum } from "./wallets/mocks/@metamask/detect-provider";
-
-jest.mock("@metamask/detect-provider", () => async () => Promise.resolve(ethereum));
-
```
mrnerdhair, line 27: This was mocking `@metamask/detect-provider` during every wallet's integration test, not just MM.

integration/src/wallets/__mocks__/@metamask/detect-provider.ts
```diff
@@ -0,0 +1,24 @@
...
+        case "eth_sendTransaction":
+          const [{ to }] = params;
+
+          return `txHash-${to}`;
+        default:
+          throw new Error(`ethereum: Unkown method ${method}`);
+      }
+    })
+  }
+}
```
mrnerdhair, line 24: This is moved into the correct directory for jest's automock to work.

integration/src/wallets/metamask.ts
```diff
@@ -10,8 +12,9 @@ export function createInfo(): core.HDWalletInfo {
 }
 
 export async function createWallet(): Promise<core.HDWallet> {
-  const wallet = new metamask.MetaMaskHDWallet();
-  await wallet.initialize();
+  const keyring = new core.Keyring();
+  const adapter = metamask.MetaMaskAdapter.useKeyring(keyring);
+  const wallet = await adapter.pairDevice()
```
mrnerdhair, line 17: Creating the wallet via the adapter (which is the public interface) now causes the provider to be pulled in from the `@metamask/detect-provider` mock.

packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -77,14 +76,11 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
...
-    try {
-      this.provider = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
-    } catch (e) {
-      console.error(e);
-    }
-
-    return Promise.resolve();
+  public initialize(): never;
+  public initialize(provider: unknown): Promise<any>;
+  public async initialize(provider?: unknown): Promise<any> {
```
pastaghost, line 81: If `provider` is a required argument, what's the motivation to mark it as optional in the function definition?

pastaghost (approved):

packages/hdwallet-metamask/src/metamask.ts
mrnerdhair, line 81:
```suggestion
  public async initialize(provider?: unknown): Promise<any> {
    if (this.provider) return;
```

mrnerdhair:
@cjthompson @pastaghost I've updated the implementation here based on chris's feedback that web's using `wallet.initialize()`. Also updated #423 to harmonize xdefi with this approach: https://github.com/shapeshift/hdwallet/pull/423/commits/cad67f970373cd435e7eb2249da2852e75da3eb4

cjthompson (approved):

PR #419, opened 2022-02-12
use async functions for portis/metamask

mrnerdhair:
this modernizes `hdwallet-portis` and `hdwallet-metamask` to use async functions instead of `return Promise.resolve()` everywhere.


0xApotheosis (approved):
Nice.

gomesalexandre (approved):
Since async/await is a syntactic sugar, you just earned a 🍬

PR #420, opened 2022-02-12
dry out describe path functions

mrnerdhair:
This moves all path description functions into core and removes all the not-quite-identical duplicates in individual wallet packages.

Edit: now rebased on #485.


0xean:
Closing as this has become overly stale at this point.

PR #418, opened 2022-02-12
return actual types from adapter.pairDevice()

mrnerdhair:

PR #423, opened 2022-02-12
minor xdefi improvements

mrnerdhair:
Requires #418 (not really, but otherwise I have to add a typecast and that would suck)

- Fix some typos
- Import less of lodash
- Improve integration test mocks
- Remove unnecessary `create()`/`info()` functions before anyone can get around to using them


packages/hdwallet-keepkey/src/adapter.ts
```diff
@@ -153,8 +153,8 @@ export class Adapter<DelegateType extends AdapterDelegate<any>> {
     return this.pairRawDevice(device, tryDebugLink);
   }
 
-  async pairRawDevice(device: DeviceType<DelegateType>, tryDebugLink?: boolean): Promise<core.HDWallet> {
+  async pairRawDevice(device: DeviceType<DelegateType>, tryDebugLink?: boolean): Promise<KeepKeyHDWallet> {
     await this.initialize([device], tryDebugLink, true);
-    return core.mustBeDefined(this.keyring.get((await this.inspectDevice(device)).serialNumber));
+    return core.mustBeDefined(this.keyring.get((await this.inspectDevice(device)).serialNumber) as KeepKeyHDWallet);
```
0xApotheosis, line 158: Should we use `get`s generic type for consistency with the rest of the codebase?

```suggestion
    return core.mustBeDefined(this.keyring.get<KeepKeyHDWallet>((await this.inspectDevice(device)).serialNumber));
```

0xApotheosis (approved):

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -11,7 +11,7 @@ class XDeFiTransport extends core.Transport {
 }
 
 export function isXDeFi(wallet: core.HDWallet): wallet is XDeFiHDWallet {
-  return _.isObject(wallet) && (wallet as any)._isXDeFi;
+  return isObject(wallet) && (wallet as any)._isXDeFi;
```
gomesalexandre, line 14: Nitpick: If we're going to import some lodash util, might as well use `_.has()` which will handle the non-object cases

integration/src/wallets/xdefi.ts
```diff
@@ -1,19 +1,106 @@
...
 
 export function name(): string {
   return "XDeFi";
 }
 
 export async function createWallet(): Promise<core.HDWallet> {
-  const wallet = createMockWallet();
-  if (!wallet) throw new Error("No XDeFi wallet found");
+  // mock xdefi
+  (globalThis as any).xfi = {
```
gomesalexandre, line 50: `packages/hdwallet-xdefi/src/adapter.ts` has the only occurrence of `globalthis as any` in the codebase currently - don't we have any way to avoid these two `any`s?

gomesalexandre (approved):

packages/hdwallet-xdefi/src/xdefi.ts
mrnerdhair: Good point -- though IMO consistency with the rest of the wallet's implementation of this type of guard is valuable, and the endgame for them should just be to nuke them wholesale (see #466 / #435).

PR #424, opened 2022-02-12
metamask: use metamask-style account path derivation

mrnerdhair:
Requires #420. Has no immediate practical impact since nothing that actually uses `hdwallet-metamask` supports multiple accounts yet.


0xean:
closing for now due to lack of need for this functionality vs risk of changes.

PR #425, opened 2022-02-12
remove some obsolete apis

mrnerdhair:
Requires #418.

- removes `Bech32` script type, which did the same thing as `SpendWitness`
- removes the `DebugLinkWallet` abstraction; #418 passes the precise type of wallet through so there's no need to introspect it to find out if it's got support


mrnerdhair:
fine, i'll break these out separately, get off my back already

PR #426, opened 2022-02-12
remove `DebugLinkWallet`

mrnerdhair:
Requires #418.

Removes the `DebugLinkWallet` abstraction; #418 passes the precise type of wallet through so there's no need to introspect it to find out if it's got support anymore.


mrnerdhair:
Hmm.. needs update with the contents of d6650a4acd90c6983eddc3df333fcc35d716c6ba. Back to draft I guess.

mrnerdhair:
Nope, I lied, that was already merged in #367. Back to non-draft.

mrnerdhair:
checked axiom and platform-shared; they don't use this anywhere. looks to me like it was only ever actually used in the integration tests, which would make it an internal implementation detail.

PR #427, opened 2022-02-12
remove `BTC(Input|Output)ScriptType.Bech32`

mrnerdhair:
Removes the `Bech32` script type, which just does exactly the same thing as `SpendWitness`.

### DIFF
--- a/integration/src/wallets/native.ts
+++ b/integration/src/wallets/native.ts
@@ -333,7 +333,7 @@ export function selfTest(get: () => core.HDWallet): void {
         coin: "Litecoin",
         isKnown: true,
         scriptType: "p2wpkh",
-        verbose: "Litecoin Account #4 (Segwit)",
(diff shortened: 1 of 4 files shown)


0xApotheosis (approved):

cjthompson (approved):
Are there any backward compatibility issues with this? Is `Bech32` currently exported? Any idea if any existing or legacy projects use that enum value?

mrnerdhair:
> Are there any backward compatibility issues with this? Is `Bech32` currently exported? Any idea if any existing or legacy projects use that enum value?

So this is an interesting question. `Bech32` is exported, but I looked through axiom and platform-shared, and it doesn't look like they use the export anywhere. Axiom's got its own array of possible `scriptType` values, which include plenty of other things that aren't on this list too, like `eth` and a bunch of other coin names, and that does have a `bech32` literal but it's used with Osmosis only. That means it should never end up used in any bitcoin context.

PR #447, opened 2022-03-04
add optional auth header to bridge

BitHighlander:
This adds an optional auth header to the tcp bridge. allows bridge applications to verify what applications are communicating with the device.

### DIFF
--- a/packages/hdwallet-keepkey-tcp/src/transport.ts
+++ b/packages/hdwallet-keepkey-tcp/src/transport.ts
@@ -12,7 +12,7 @@ export class TransportDelegate implements keepkey.TransportDelegate {
     this.config = {
       ...config,
       headers: {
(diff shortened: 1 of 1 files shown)


packages/hdwallet-keepkey-tcp/src/transport.ts
```diff
@@ -12,6 +12,7 @@ export class TransportDelegate implements keepkey.TransportDelegate {
     this.config = {
       ...config,
       headers: {
+        'Authorization': config?.headers?.serviceKey || '',
```
mrnerdhair, line 15: This will already be included if you set `config.headers.Authorization`... not sure there's any value in including it separately here. If we do, though, we should add it to the `Config` type explicitly.

mrnerdhair (changes requested):

Issue #445, opened 2022-03-01
Feature: Create a "demo" wallet type


cjthompson:
# Description
A crypto website wants a user to be able to use their site while not requiring them to pair their own wallet. 

Benefits:
* Allows a user to see all the features of the site.
* Allows for creation of automated tests against the website used a known wallet state

AC:
* A new, separate wallet module (e.g., `hdwallet-demo`)
* The module cannot accept a private key of any kind
* `getPublicKeys()` must return a fixed/static list of public keys (these could be hard-coded or passed into the constructor)
* functions that interact with the blockchain should return mock data where appropriate (such as broadcast tx)

BitHighlander:
>functions that interact with the blockchain should return mock data where appropriate (such as broadcast tx)

wouldn't that have to be done in the chain-adapters? hdwallet has no blockchain interactions.

Is the goal to show a user valid addresses/pubkeys? because that could introduce a users actually sending money into it on accident?

if its fake data, then that is going to need a lot of validator issues.

imho, this isn't really a hdwallet bounty, its more of a web project. and its going to get rather in-depth if we want to UX to submit/send/verify tx's and act identical to a live application.

cjthompson:
> > functions that interact with the blockchain should return mock data where appropriate (such as broadcast tx)
> 
> wouldn't that have to be done in the chain-adapters? hdwallet has no blockchain interactions.

Sorry that should say "interact with the private key".

> imho, this isn't really a hdwallet bounty, its more of a web project. and its going to get rather in-depth if we want to UX to submit/send/verify tx's and act identical to a live application.

The web project leverages an hdwallet instance everywhere. We want to try to do a "proof of concept" to see if we can leverage that to make some form of demo wallet.

Whoever works on this is expected, at this time, to only build a proof of concept to see if this is a feasible approach. If during the building of the POC we discover that there are significant blockers or hurtles to this approach, then the recommendation can be made that we need to find a different approach to the demo wallet.

If it is feasible, then at that time would be put more time into figuring out how we would make a full implementation in web, which could include changes to `ChainAdapters` if necessary.

PR #452, opened 2022-03-08
native: support dummy pubkey-only wallets

mrnerdhair:
In support of #445, this adds a new isolation engine to the native wallet that doesn't actually have any private keys. Instead, it takes a list of xpubs, parses them, and uses their embedded parent fingerprints to put them into a tree. Operations which require a private key (signing, ECDH, and deriving hardened child keys whose xpubs aren't in the list of preloaded public key material) throw a DummyEngineError.


packages/hdwallet-native/src/crypto/isolation/engines/dummy/types.ts
```diff
@@ -0,0 +1,22 @@
+import { ChainCode } from "../../core/bip32";
+import { CompressedPoint } from "../../core/secp256k1";
+
+export class DummyEngineError extends Error {
+  constructor() {
+    super("the dummy isolation engine can't do this because it doesn't have any actual private keys");
```
cjthompson, line 6: improve error message

packages/hdwallet-native/src/crypto/isolation/engines/dummy/bip39.ts
```diff
@@ -0,0 +1,71 @@
...
+    readonly xpubTree: ParsedXpubTree;
+
+    protected constructor(xpubTree: ParsedXpubTree) {
+        this.xpubTree = xpubTree;
+    }
+
+    static async create(xpubList: string): Promise<BIP39.Mnemonic> {
+        const parsedXpubs: ParsedXpubTree[] = xpubList.split(" ").map(xpub => {
+            const xpubBuf = bs58check.decode(xpub);
+            if (xpubBuf.length !== 78) throw new Error("bad xpub: ")
```
cjthompson, line 25:
```
throw new Error('Isolation:Engine:Dummy:bip39:create - Invalid xpub')
```

cjthompson (approved):

packages/hdwallet-native/src/crypto/isolation/engines/dummy/bip39.ts
mrnerdhair:
```suggestion
            if (xpubBuf.length !== 78) throw new Error("Isolation.Engine.Dummy.BIP39.create - Invalid xpub")
```

PR #477, opened 2022-03-15
Make compatible with EIP 1559

ErvinDimitri:


packages/hdwallet-core/src/ethereum.ts
```diff
@@ -33,7 +33,7 @@ export interface ETHGetAddress {
 
 export interface ETHSignTx {
   /** bip32 path to sign the transaction from */
-  addressNList: BIP32Path;
+  addressNList?: BIP32Path;
```
mrnerdhair, line 36: This can't be optional b/c it's what's used to tell the wallet which key to sign with.

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -58,6 +58,8 @@ export interface ETHSignTx {
    * Device must `ethSupportsNativeShapeShift()`
    */
   exchangeType?: ExchangeType;
+  type: string,
+  accessList?:[]
```
mrnerdhair, line 62: Using a `string` here isn't a great choice, and from the field name I can't even tell what you mean! If you're talking about the EIP-2718 transaction type, that's at a lower level of abstraction than is appropriate here. The correct TX type is inferred from whether `gasPrice` or `maxFeePerGas`/`maxPriorityFeePerGas` is present; including an explicit flag for it would make the type denormalized.

Happy to support `accessList` here -- though that's really EIP-2930, not EIP-1559 -- but that type `[]` definitely isn't correct. Maybe `[string, string[]][]` with a doc comment indicating that those refer to addresses / storage keys respectively? In any case, we'll need to see an implementation in at least `hdwallet-native` as well as accompanying unit and integration tests, to move forward.

mrnerdhair (changes requested):
This isn't going to get merged as-is, but the adding EIP-2930 support would be worthwhile. I'd recommend opening an issue about it for traceability.

packages/hdwallet-core/src/ethereum.ts
ErvinDimitri: I appreciate the feedback and I will make some improvements.

PR #479, opened 2022-03-16
fix: strict gas price specification in ETHSignTx

### LINKED ISSUE #478: `ETHSignTx` type has ambiguous gas price
The `ETHSignTx` type currently marks `gasPrice`, `maxFeePerGas`, and `maxPriorityFeePerGas` all as optional. This means it matches TXes with no gas info, as well as TXes which have both gas fee types specified -- and are thus ambiguous about whether they should be run as legacy or EIP-1559 TXes. The caller must be able to rely on a certain method will being used for accurate sendMax calculations.

- Exactly one of `gasPrice` or a pair of `maxFeePerGas` and `maxPriorityFeePerGas` should be allowed.

mrnerdhair:
Fixes #478. Compatble with, but not required for, https://github.com/shapeshift/lib/pull/440 and https://github.com/shapeshift/web/pull/1182.

### DIFF
--- a/integration/src/ethereum/ethereum.ts
+++ b/integration/src/ethereum/ethereum.ts
@@ -199,38 +199,6 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
     );
 
     test(
-      "ethSignTx() - ETH EIP-1559 (optional)",
-      async () => {
-        if (!wallet) return;
-
-        if (!(await wallet.ethSupportsEIP1559())) {
-          return;
-        }
-
-        const res = await wallet.ethSignTx({
-          addressNList: core.bip32ToAddressNList("m/44'/60'/0'/0/0"),
-          nonce: "0x01",
-          gasPrice: "0x1dcd65000",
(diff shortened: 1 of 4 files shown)


integration/src/ethereum/ethereum.ts
```diff
@@ -203,38 +203,6 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
       TIMEOUT
     );
 
-    test(
```
pastaghost, line 206: I agree with the `ETHSignTx` type change below, but we should have separate tests for both tx formats (EIP-1559 and legacy fee parameters)

mrnerdhair: we've got them in the form of "ethSignTx() - ETH EIP-1559" above and "ethSignTx() - ERC20" below; this test was specifically targeting providing the combination of both types of parameters, which is now illegal

pastaghost (approved):

cjthompson (approved):

Issue #472, opened 2022-03-15
`.gitignore` should be a whitelist


mrnerdhair:
### Overview

`.gitignore` is currently allow-by-default, which inevitably leads to it being cluttered and difficult to maintain. Having a deny-by-default `.gitignore` would be preferable.

### Reference

A deny-by-default `.gitignore` looks something like [this](https://github.com/shapeshift/proto-tx-builder/blob/main/.gitignore).

### Acceptance Criteria

- Replace the `.gitignore` file with a deny-by-default version.
- The new `.gitignore` file should not ignore any files not currently ignored.
- The new `.gitignore` file should ignore at least all the files ignored by the current `.gitignore` which are present after a fresh `git clone` followed by `yarn install`, `yarn build`, `yarn test`, and `yarn test:integration`.

### Ownership

- [X] As the sponsor of this bounty I will review the changes in a preview environment (ops/product) or review the PR (engineering)

### Bounty Hunters

- [Join our Discord](https://discord.gg/shapeshift)
- Include an expected timeline for you to complete work in the work plan when you apply for this bounty!
- Please refer to [this link](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters) for some basic info

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 264 years, 8 months from now.
Please review their action plans below:


**1) [sdmg15](https://gitcoin.co/sdmg15) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/shapeshift/hdwallet/472/100028619?mutate_worker_action=approve&worker=sdmg15) | [reject worker](https://gitcoin.co/issue/shapeshift/hdwallet/472/100028619?mutate_worker_action=reject&worker=sdmg15))_.**

I've read the instructions and I'll add the deny-by-default .gitignore required
**2) [lolboysg](https://gitcoin.co/lolboysg) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/shapeshift/hdwallet/472/100028619?mutate_worker_action=approve&worker=lolboysg) | [reject worker](https://gitcoin.co/issue/shapeshift/hdwallet/472/100028619?mutate_worker_action=reject&worker=lolboysg))_.**

I've already worked with the codebase so I am going to change the .gitignore for a deny by default one
[...]

PR #480, opened 2022-03-16
chore: .gitignore should be a whitelist

thengoster:
Pr for issue #472

Replaced .gitignore with deny-by-default version by first ignoring everything, and then whitelisting the directories/files we want in our git repository.


.gitignore
```diff
@@ -1,21 +1,54 @@
...
+!/SECURITY.md
+!/SUMMARY.md
+!/changelog.md
+!/code-of-conduct.md
+!/jest.config.js
+!/lerna.json
+!/package.json
+!/tsconfig.json
+!/vercel.json
+!/yarn.lock
```
mrnerdhair, line 54:
```suggestion
!/yarn.lock

```

.gitignore
```diff
@@ -1,21 +1,54 @@
...
+!/integration/src/**
+
+!/packages/
+!/packages/*
+!/packages/**/src/
+!/packages/**/src/**
+!/packages/**/.npmignore
+!/packages/**/jest.config.js
+!/packages/**/package.json
+!/packages/**/tsconfig.json
```
mrnerdhair, line 29:
```suggestion
!/packages/*/src/
!/packages/*/src/**
!/packages/*/.npmignore
!/packages/*/jest.config.js
!/packages/*/package.json
!/packages/*/tsconfig.json
```
(haven't tested this yet, but we probably don't want to apply these recursively)

mrnerdhair (approved):
Personally I think this is better sorted, but I'll leave it up to reviewer number 2; I also made .prettierignore a symlink so we don't have to worry about it getting out of sync anymore.

pastaghost:
LGTM. I just suggest that we update the README with some text for future developers indicating that newly added root-level files/subdirectories won't be tracked by default and exceptions will need to be added to the .gitignore.

pastaghost (approved):

PR #453, opened 2022-03-09
[WIP] feat: add eslint

NeOMakinG:
## Informations

This PR introduces ESLint in the `hdwallet` repository.

As the project has a lot of legacy, we decided to ignore a lot of errors that could be fixed later or while refactoring. 

Most of ignored are scoped the the rule failing, so it permits to detect some others problems. 

Closes #353

PR #485, opened 2022-03-22
feat: add-eslint

mrnerdhair:
Supercedes #453. It's the same thing, done by @NeOMakinG, but rebased for easier review. Needed its own PR since I rewrote the history.


integration/src/secret/secret.ts
```diff
@@ -51,7 +51,7 @@ export function secretTests(get: () => { wallet: core.HDWallet; info: core.HDWal
             path: core.bip32ToAddressNList("m/44'/529'/0'/0/0"),
             coin: "Secret",
           })
-        );
+        ).toMatchInlineSnapshot();
```
NeOMakinG, line 54: I think we should generate it!

NeOMakinG:
@mrnerdhair just a little thing that I'll try to push today, then let's open it for review?

integration/src/secret/secret.ts
NeOMakinG: Ok I get it, it doesn't return anything, we don't care then

mrnerdhair:
Rebased, updated, and ready for review!

cjthompson (approved):

PR #493, opened 2022-03-24
fix: lowercase checksummed address in ethGetAddress()

gomesalexandre:
## Description

This fixes addresses being returned as checksum addresses by Portis, KeepKey, and Native wallets.

This is a discrepancy from Metamask hdwallet, which methods return the lowercase version of an address, and as a result:

- There is a visual discrepancy across the app in how addresses are displayed, depending on the used wallet and where in the app you are
- When getting the CAIP19 for an address, we get the CAIP19 for the cheksummed version, which is the correct behavior since the `asset_id`/`asset_type` parts are case-sensitive ([docs](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md#syntax)). 
Because of this, we are getting different CAIP19s across web which results in a bug when trying to compare account specifiers with the current account ID:

https://github.com/shapeshift/web/blob/ad806dee305535434440679906f1d74f130a8fa4/src/pages/Accounts/AccountToken/AccountToken.tsx#L28

### Risk

I fixed this at the root level rather than throwing `.toLowerCase()` everywhere.

There is a risk that lowercase addresses are not supported everywhere, but since e.g the Metamask hdwallet uses these (by making an ethereum JsonRPC [`eth_account`](https://eth.wiki/json-rpc/API#eth_accounts) call which returns the lowercased version, that is supported in unchained) this shouldn't be an issue.

@mrnerdhair WDYT, do we want to keep the chesksum addresses in hdwallet since it's a lower-level package, or is it fine to lowercase them? If we want to keep them as is, we will:

1. Need to find a way to implement the same behavior in metamask hdwallet for consistency, since the JsonRPC endpoint gives us the lowercase version
2. Need to move the lowercasing logic to chain-adapters, or one level more, in web.

### Testing

No testing steps here until the dependency is bumped in `chain-adapters` and the `chain-adapters` dependency is bumped in `web`.

### DIFF
--- a/packages/hdwallet-keepkey/src/ethereum.ts
+++ b/packages/hdwallet-keepkey/src/ethereum.ts
@@ -193,7 +193,7 @@ export async function ethGetAddress(transport: Transport, msg: core.ETHGetAddres
   else if (ethAddress.hasAddress()) address = "0x" + core.toHexString(ethAddress.getAddress_asU8());
   else throw new Error("Unable to obtain ETH address from device.");
 
-  return address;
+  return address.toLowerCase();
 }
 
 export async function ethSignMessage(transport: Transport, msg: core.ETHSignMessage): Promise<core.ETHSignedMessage> {
--- a/packages/hdwallet-native/src/ethereum.test.ts
+++ b/packages/hdwallet-native/src/ethereum.test.ts
@@ -48,7 +48,7 @@ describe("NativeETHWallet", () => {
 
   it("should generate a correct ethereum address", async () => {
     expect(await wallet.ethGetAddress({ addressNList: core.bip32ToAddressNList("m/44'/60'/0'/0/0") })).toBe(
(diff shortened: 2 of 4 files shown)


gomesalexandre:
Agreed with @0xdef1cafe not to solve this in hdwallet, since it should give us some reliable data:
> information theory wise we want to retain as much information as possible for as long as possible

Any additional parsing should be done on consumers of hdwallet.

mrnerdhair:
We should fix this, but the fix is to return checksummed addresses from hdwallet everywhere. Clients can call `toLowerCase()` if they want to, but generating the checksum requires special, [potentially chain-specific](https://eips.ethereum.org/EIPS/eip-1191) knowledge and should be well encapsulated instead of forcing the client to import `web3` or something to fix it themselves.

mrnerdhair:
(Note to self: make issue for this)

PR #508, opened 2022-04-10
remove unused adapter.initialize() return value from the public API

mrnerdhair:
Neither webv2 nor axiom uses the return value from `adapter.initialize()`; removing it from the API would reduce the complexity of the boilerplate new wallet implementations (xdefi, tally, keplr, etc) have to cut-n-paste.

(This is a suggestion to change a core part of the API. It should not be merged until consensus is reached on if it's a good idea or not, so it's a draft until we have that conversation.)


0xean:
closing as stale draft PR

PR #509, opened 2022-04-10
chore: remove unused `currentDeviceID` field from `MetaMaskAdapter`

mrnerdhair:
This field was inherited from the copy-n-paste of `PortisAdapter`. It served a purpose there -- namely, `PortisAdapter` would detect a user logging out from Portis and remove the wallet from the keyring -- but it doesn't do anything in `MetaMaskAdapter`.


gomesalexandre (approved):
LGTM. Double-checked and this isn't used anywhere in the module, as well as other projects (e.g chain-adapters, web).

cjthompson (approved):

PR #516, opened 2022-04-18
feat: lazy loading for dynamic imports

mrnerdhair:
The KeepKey updater can't use a mainline copy of hdwallet because some of the cosmos deps of `proto-tx-builder` break on import under Node 14.x. (Specifically, they need `bigint` -- and none of the polyfills *quite* cut it.) Of course, the cosmos code isn't actually used at all in the updater -- so lazy-loading it is a cheap way to work around the issue, and it'll give us some extra page-load performance on web to boot.


cjthompson:
`node` v14 definitely supports BigInt.

```
> node
Welcome to Node.js v14.19.0.
Type ".help" for more information.
> BigInt
[Function: BigInt]
> var a = 1n
undefined
> a * 2n
2n
> 
```

packages/hdwallet-native/src/binance.ts
```diff
@@ -1,13 +1,16 @@
...
-import type * as bnbSdk from "bnb-javascript-sdk-nobroadcast";
+import type * as bnbSdkTypes from "bnb-javascript-sdk-nobroadcast";
 import CryptoJS from "crypto-js";
+import PLazy from "p-lazy";
 
 import * as Isolation from "./crypto/isolation";
 import { NativeHDWalletBase } from "./native";
 import * as util from "./util";
 
+const bnbSdk = PLazy.from(async () => await import("bnb-javascript-sdk-nobroadcast"));
```
cjthompson, line 12: I don't understand this. What's wrong with:
```suggestion
const bnbSdk = PLazy.from(() => import("bnb-javascript-sdk-nobroadcast"));
```

packages/hdwallet-native/src/binance.ts
```diff
@@ -88,8 +91,7 @@ export function MixinNativeBinanceWallet<TBase extends core.Constructor<NativeHD
         if (!tx.sequence) tx.sequence = "0";
         if (!tx.source) tx.source = "0";
 
-        const bnbSdk = await import("bnb-javascript-sdk-nobroadcast");
```
cjthompson, line 91: isn't this import already lazy because it doesn't get called until someone calls `binanceSignTx`?

packages/hdwallet-native/src/crypto/isolation/adapters/bip32.ts
```diff
@@ -47,7 +48,7 @@ export class BIP32Adapter extends ECPairAdapter implements BIP32InterfaceAsync {
     networkOrParent?: BIP32Adapter | Network,
     index?: number
   ): Promise<BIP32Adapter> {
-    await btccryptoReady;
+    await Promise.all([btccryptoReady, ECPairAdapter.prepare()]);
```
cjthompson, line 51: How is this change related to lazy loading?

Can you add a comment explaining _why_ `prepare()` needs to be called here?

packages/hdwallet-native/src/crypto/isolation/adapters/bitcoin.ts
```diff
@@ -23,6 +24,10 @@ export class ECPairAdapter implements SignerAsync, ECPairInterfaceAsync {
     this._network = network;
   }
 
+  protected static async prepare() {
+    await networksReady;
+  }
```
cjthompson, line 29: I don't like this pattern. There's no documentation explaining why `prepare` exists and it's unclear to a developer that `prepare` must be called before they can read from `networks`.

cjthompson (approved):
approved with minor questions

mrnerdhair:
> `node` v14 definitely supports BigInt.

Huh! Must be whichever flavor of node the particular electron version's shipping then.

packages/hdwallet-native/src/binance.ts
mrnerdhair: Yeah, just following the same pattern as the others.

packages/hdwallet-native/src/binance.ts
mrnerdhair: nothing in particular, I just have a general aversion to returning promises without awaiting them; makes for better stack traces imo.

packages/hdwallet-native/src/crypto/isolation/adapters/bitcoin.ts
mrnerdhair: I don't either. `create()` is a hack in the first place, I wish you could do e.g. `async constructor(): Promise<this> { await super() }`. If there's a better pattern we should look into it.

In the meantime, I've added extra comments as a stopgap.

packages/hdwallet-native/src/crypto/isolation/adapters/bip32.ts
mrnerdhair: Extra comments added; also see https://github.com/shapeshift/hdwallet/pull/516#discussion_r852466460.

packages/hdwallet-native/src/binance.ts
mrnerdhair: applied your suggestion on all similar lines.

0xdef1cafe (approved):

PR #522, opened 2022-04-21
fix: updated intanceof checks to allow Buffer intances

theoboldfrazier:
Due to polyfill of Buffer on webv2, the function `fromB64ToArray` was returning type Buffer, instead of Uint8Array. This was causing an error to be thrown and blocking decryption of a wallet.

This PR updates the checks to allow instanceof Buffer in addition to the existing ArrayBuffer and Uint8Array types.


mrnerdhair (changes requested):
Pretty sure this is caused by jest pulling prototype shenanigans with Uint8Array which break `instanceof`. We should fix jest, or at most the return type of `fromB64ToArray()`, instead of widening the runtime checks.

mrnerdhair:
Double-checking; is this issue still relevant / can we close this?

cjthompson:
It seems to me that `Buffer` is a valid input to those functions, seeing as how no code changes were needed to support `Buffer`. I don't see a downside here.

packages/hdwallet-native/src/crypto/utils.ts
```diff
@@ -12,6 +12,6 @@ export function fromBufferToUtf8(buffer: ArrayBuffer): string {
   return new TextDecoder().decode(buffer);
 }
 
-export function fromB64ToArray(str: string): Uint8Array {
+export function fromB64ToArray(str: string): Uint8Array | Buffer {
```
0xApotheosis, line 15: Doesn't `Buffer.from()` always return `Buffer` for all of its overloads?

0xean:
closing as stale.

PR #538, opened 2022-06-09
feat: make the bnb sdk an optionalDependency

mrnerdhair:
- We don't use BNB support in shapeshift/web at the moment, and having to ship a heavy SDK we don't use is silly
- The old `bnb-javascript-sdk-nobroadcast` is very stale and isn't maintained under the org (I made `@shapeshiftoss/bnb-javascript-sdk` to fix that, it's got CI and everything)
- The BNB SDK version includes various transitive dependencies which produce `npm audit` warnings. None of these are applicable to our specific use of it, but providing a way for other devs not to have to depend on this package -- and therefore get bugged by their users -- would be a nice-to-have, especially if they only want e.g. Ethereum support in the first place.


0xean:
closing as stale.

PR #540, opened 2022-06-10
feat: add ethGetChainId method

gomesalexandre:
### Description
This adds an `ethGetChainId` method to `hdwallet-metamask`, `hdwallet-xdefi` and `hdwallet-tallyho` to be able to detect the current `chainId`.

Note: We currently are able to call `provider.chainId` and get the right chainId. It is working for MetaMask, XDEFI and Tally currently, however, this is both a deprecated as well as a non-standard property and we shouldn't be relying on it for all wallets that support ethereum JSONRPC calls: https://docs.metamask.io/guide/ethereum-provider.html#legacy-api

### Issue

Partially tackles https://github.com/shapeshift/web/issues/1843

#### MetaMask:
 
Ethereum Mainnet: 

image

Polygon Mainnet:

image

Gnosis Chain:

image

### Tally 

Ethereum Mainnet:

image

Note: Tally currently only supports mainnet, but from what I can see on their codebase, other EVM networks supports is already partially implemented so it makes sense to add this for Tally as well. 

### XDEFI

Ethereum Mainnet:

image

Polygon Mainnet:

image

### DIFF
--- a/packages/hdwallet-core/src/ethereum.ts
+++ b/packages/hdwallet-core/src/ethereum.ts
@@ -111,7 +111,7 @@ export interface ETHWalletInfo extends HDWalletInfo {
    * Get the current chainId from ethereum's JSON RPC
    * https://eips.ethereum.org/EIPS/eip-695
    */
-  ethGetChainId?(): Promise<number>;
+  ethGetChainId?(): Promise<number | null>;
 
   /**
    * Does the device support internal transfers without the user needing to
--- a/packages/hdwallet-core/src/ethereum.ts
+++ b/packages/hdwallet-core/src/ethereum.ts
@@ -108,6 +108,12 @@ export interface ETHWalletInfo extends HDWalletInfo {
   ethSupportsNetwork(chain_id: number): Promise<boolean>;
 
   /**
+   * Get the current chainId from ethereum's JSON RPC
+   * https://eips.ethereum.org/EIPS/eip-695
+   */
+  ethGetChainId?(): Promise<number | void>;
+
+  /**
    * Does the device support internal transfers without the user needing to
    * confirm the destination address?
    */
--- a/packages/hdwallet-core/src/ethereum.ts
+++ b/packages/hdwallet-core/src/ethereum.ts
@@ -111,7 +111,7 @@ export interface ETHWalletInfo extends HDWalletInfo {
    * Get the current chainId from ethereum's JSON RPC
    * https://eips.ethereum.org/EIPS/eip-695
    */
-  ethGetChainId?(): Promise<number | void>;
+  ethGetChainId?(): Promise<number>;
 
   /**
    * Does the device support internal transfers without the user needing to
--- a/packages/hdwallet-metamask/src/metamask.ts
+++ b/packages/hdwallet-metamask/src/metamask.ts
@@ -250,12 +250,16 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
     return chainId === 1;
   }
 
-  public async ethGetChainId(): Promise<number> {
-    // at this point, we know that we're in the context of a valid MetaMask provider
-    const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
-    // chainId as hex string
-    const chainId: string = await provider.request({ method: "eth_chainId" });
-    return parseInt(chainId, 16);
+  public async ethGetChainId(): Promise<number | null> {
+    try {
+      const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
+      // chainId as hex string
+      const chainId: string = await provider.request({ method: "eth_chainId" });
+      return parseInt(chainId, 16);
+    } catch (e) {
+      console.error(e);
+      return null;
+    }
   }
 
   public async ethSupportsSecureTransfer(): Promise<boolean> {
--- a/packages/hdwallet-metamask/src/metamask.ts
+++ b/packages/hdwallet-metamask/src/metamask.ts
@@ -1,3 +1,4 @@
+import detectEthereumProvider from "@metamask/detect-provider";
 import * as core from "@shapeshiftoss/hdwallet-core";
 import _ from "lodash";
 
@@ -70,6 +71,16 @@ export class MetaMaskHDWalletInfo implements core.HDWalletInfo, core.ETHWalletIn
 
   public async ethSupportsNetwork(chainId = 1): Promise<boolean> {
     return chainId === 1;
+  }
+
+  public async ethGetChainId(): Promise<number> {
+    const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
+    if (!provider) {
+      throw new Error("Cannot get chainId");
+    }
+    // chainId as hex string
+    const chainId: string = provider.request({ method: "eth_chainId" });
+    return parseInt(chainId, 16);
   }
 
   public async ethSupportsSecureTransfer(): Promise<boolean> {
@@ -247,6 +258,16 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
 
   public async ethSupportsNetwork(chainId = 1): Promise<boolean> {
     return chainId === 1;
+  }
+
+  public async ethGetChainId(): Promise<number> {
+    const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
+    if (!provider) {
+      throw new Error("Cannot get chainId");
+    }
+    // chainId as hex string
+    const chainId: string = provider.request({ method: "eth_chainId" });
+    return parseInt(chainId, 16);
   }
 
   public async ethSupportsSecureTransfer(): Promise<boolean> {
--- a/packages/hdwallet-metamask/src/metamask.ts
+++ b/packages/hdwallet-metamask/src/metamask.ts
@@ -74,10 +74,8 @@ export class MetaMaskHDWalletInfo implements core.HDWalletInfo, core.ETHWalletIn
   }
 
   public async ethGetChainId(): Promise<number> {
+    // at this point, we know that we're in the context of a valid MetaMask provider
     const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
-    if (!provider) {
-      throw new Error("Cannot get chainId");
-    }
     // chainId as hex string
     const chainId: string = await provider.request({ method: "eth_chainId" });
     return parseInt(chainId, 16);
@@ -261,10 +259,8 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
   }
 
   public async ethGetChainId(): Promise<number> {
+    // at this point, we know that we're in the context of a valid MetaMask provider
     const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
-    if (!provider) {
-      throw new Error("Cannot get chainId");
-    }
     // chainId as hex string
     const chainId: string = await provider.request({ method: "eth_chainId" });
     return parseInt(chainId, 16);
--- a/packages/hdwallet-metamask/src/metamask.ts
+++ b/packages/hdwallet-metamask/src/metamask.ts
@@ -1,4 +1,3 @@
-import detectEthereumProvider from "@metamask/detect-provider";
 import * as core from "@shapeshiftoss/hdwallet-core";
 import _ from "lodash";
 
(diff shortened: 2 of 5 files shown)


packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -72,6 +73,16 @@ export class MetaMaskHDWalletInfo implements core.HDWalletInfo, core.ETHWalletIn
     return chainId === 1;
   }
 
+  public async ethGetChainId(): Promise<number> {
+    const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
+    if (!provider) {
+      throw new Error("Cannot get chainId");
+    }
+    // chainId as hex string
```
gomesalexandre, line 81: Note: Web currently checks for the hex string, but now that we have a proper method for getting the chainId and we're just accessing the `provider.chainId` property, there's probably no reason to keep it as hex string anymore.

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -107,6 +107,12 @@ export interface ETHWalletInfo extends HDWalletInfo {
    */
   ethSupportsNetwork(chain_id: number): Promise<boolean>;
 
+  /**
+   * Get the current chainId from ethereum's JSON RPC
+   * https://eips.ethereum.org/EIPS/eip-695
+   */
+  ethGetChainId?(): Promise<number | void>;
```
mrnerdhair, line 114: Or null, or whatever, but not void -- you can return anything, including a number, from a function whose type is assignable to a function whose return type is void. More to the point, though, if the method's optional already and we don't need to determine at runtime if the question is valid (which I don't think we do?) the return type shouldn't be ambiguous.

```suggestion
  ethGetChainId?(): Promise<number | undefined>;
```

mrnerdhair (dismissed):
Requesting changes here because I'm not convinced we shouldn't abstract over this distinction instead of exposing it to a higher layer. For example, we know the chain type when a transaction's built; can we check/switch to the correct chain at that point?

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -62,6 +63,53 @@ export class TallyHoHDWalletInfo implements core.HDWalletInfo, core.ETHWalletInf
     return chainId === 1;
   }
 
+  private async detectTallyProvider(): Promise<TallyHoEthereumProvider | null> {
```
gomesalexandre, line 66: Note: We might get away with a lighter implementation since the checks for being in the context of a Tally provider should be completed, but IMO doesn't hurt to err on the side of safety and keep the full detection implementation

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -214,6 +262,53 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
...
```
gomesalexandre, line 265: Note: We might get away with a lighter implementation since the checks for being in the context of a Tally provider should be completed, but IMO doesn't hurt to err on the side of safety and keep the full detection implementation

packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -72,6 +73,14 @@ export class MetaMaskHDWalletInfo implements core.HDWalletInfo, core.ETHWalletIn
...
   }
 
+  public async ethGetChainId(): Promise<number> {
+    // at this point, we know that we're in the context of a valid MetaMask provider
+    const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
+    // chainId as hex string
+    const chainId: string = await provider.request({ method: "eth_chainId" });
+    return parseInt(chainId, 16);
+  }
+
```
pastaghost, line 83: I think we only need this method in the `MetaMaskHDWallet` class.

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -62,6 +63,53 @@ export class TallyHoHDWalletInfo implements core.HDWalletInfo, core.ETHWalletInf
...
+  }
+
+  public async ethGetChainId(): Promise<number> {
+    // at this point, we know that we're in the context of a valid TallyHo provider
+    const provider: any = await this.detectTallyProvider();
+    // chainId as hex string
+    const chainId: string = await provider.request({ method: "eth_chainId" });
+    return parseInt(chainId, 16);
+  }
+
```
pastaghost, line 112: I think we only need these methods in the `TalloHoHDWallet` class.

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -63,6 +64,13 @@ export class XDEFIHDWalletInfo implements core.HDWalletInfo, core.ETHWalletInfo
     return chainId === 1;
   }
 
+  public async ethGetChainId(): Promise<number> {
+    const provider: any = await detectEthereumProvider({ mustBeMetaMask: false, silent: false, timeout: 3000 });
+    // chainId as hex string
+    const chainId: string = await provider.request({ method: "eth_chainId" });
+    return parseInt(chainId, 16);
+  }
+
```
pastaghost, line 73: I think we only need this method in the `XDEFIHDWallet` class.

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -220,6 +228,14 @@ export class XDEFIHDWallet implements core.HDWallet, core.ETHWallet {
     return chainId === 1;
   }
 
+  public async ethGetChainId(): Promise<number> {
+    // at this point, we know that we're in the context of a valid XDEFI provider
+    const provider: any = await detectEthereumProvider({ mustBeMetaMask: false, silent: false, timeout: 3000 });
+    // chainId as hex string
+    const chainId: string = await provider.request({ method: "eth_chainId" });
+    return parseInt(chainId, 16);
```
pastaghost, line 236: I'm not sure what possible failure conditions exist for the async function calls here, but maybe this should be wrapped in a `try/catch` for safety.

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -214,6 +262,53 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
...
+      }
+    });
+  }
+
+  public async ethGetChainId(): Promise<number> {
+    // at this point, we know that we're in the context of a valid TallyHo provider
+    const provider: any = await this.detectTallyProvider();
+    // chainId as hex string
+    const chainId: string = await provider.request({ method: "eth_chainId" });
+    return parseInt(chainId, 16);
```
pastaghost, line 309: I'm not sure what possible failure conditions exist for the async function calls here, but maybe this should be wrapped in a `try/catch` for safety.

packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -249,6 +258,14 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
     return chainId === 1;
   }
 
+  public async ethGetChainId(): Promise<number> {
+    // at this point, we know that we're in the context of a valid MetaMask provider
+    const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
+    // chainId as hex string
+    const chainId: string = await provider.request({ method: "eth_chainId" });
+    return parseInt(chainId, 16);
```
pastaghost, line 266: I'm not sure what possible failure conditions exist for the async function calls here, but maybe this should be wrapped in a `try/catch` for safety.

pastaghost (changes requested):

packages/hdwallet-xdefi/src/xdefi.ts
gomesalexandre: Good point, I'm not sure about this one.
Since we've narrowed the provider itself to be valid, I would think anything going wrong with calling the Ethereum JSON-RPC provider with such method is really an edge case, what do you think? How would we handle the failed call?

packages/hdwallet-tallyho/src/tallyho.ts
gomesalexandre: Good point, I'm not sure about this one.
Since we've narrowed the provider itself to be valid, I would think anything going wrong with calling the Ethereum JSON-RPC provider with such method is really an edge case, what do you think? How would we handle the failed call?

packages/hdwallet-metamask/src/metamask.ts
gomesalexandre: Good point, I'm not sure about this one.
Since we've narrowed the provider itself to be valid, I would think anything going wrong with calling the Ethereum JSON-RPC provider with such method is really an edge case, what do you think? How would we handle the failed call?

packages/hdwallet-tallyho/src/tallyho.ts
gomesalexandre: Seems to work without, removed in a53319d3ad68006635a6bba6dd5d8079c718d32d

packages/hdwallet-xdefi/src/xdefi.ts
gomesalexandre: Seems to work without, removed it in a53319d3ad68006635a6bba6dd5d8079c718d32d

packages/hdwallet-metamask/src/metamask.ts
pastaghost: The particular way we are using hdwallet in web guarantees that we'll be in the context of a valid MetaMask provider here, but hdwallet is used in other non-shapeshift projects. I'd abandon any assumptions about use case here and account for the cases where either of the await-ed calls fail. In this case, I'd throw an error, log the error to console, and change the return type to `Promise<number | null>`.

packages/hdwallet-tallyho/src/tallyho.ts
pastaghost: The particular way we are using hdwallet in web guarantees that we'll be in the context of a valid TallyHo provider here, but hdwallet is used in other non-shapeshift projects. I'd abandon any assumptions about use case here and account for the cases where either of the await-ed calls fail. In this case, I'd throw an error, log the error to console, and change the return type to `Promise<number | null>`.

packages/hdwallet-xdefi/src/xdefi.ts
pastaghost: The particular way we are using hdwallet in web guarantees that we'll be in the context of a valid XDEFI provider here, but hdwallet is used in other non-shapeshift projects. I'd abandon any assumptions about use case here and account for the cases where either of the await-ed calls fail. In this case, I'd throw an error, log the error to console, and change the return type to `Promise<number | null>`.

pastaghost (changes requested):

packages/hdwallet-xdefi/src/xdefi.ts
gomesalexandre: Added a try/catch in 75cc8e245128fe9827f15f0f3c37a98a507893e4 (note: I didn't throw since throwing would effectively make us unable to return anything else than void)

packages/hdwallet-metamask/src/metamask.ts
gomesalexandre: Added a try/catch in 75cc8e245128fe9827f15f0f3c37a98a507893e4 (note: I didn't throw since throwing would effectively make us unable to return anything else than void)

pastaghost (dismissed):

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -214,6 +215,57 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
     return chainId === 1;
   }
 
+  private async detectTallyProvider(): Promise<TallyHoEthereumProvider | null> {
```
0xApotheosis, line 218: Did we want to add the same comment that `packages/hdwallet-tallyho/src/adapter.ts` has which references the source of this code so it's clear it isn't written by us?

```ts
  /*
   * Tally works the same way as metamask.
   * This code is copied from the @metamask/detect-provider package
   * @see https://www.npmjs.com/package/@metamask/detect-provider
   */
```

0xApotheosis (approved):
Looks good, one non-blocking comment that we should probably address before merging.

packages/hdwallet-tallyho/src/tallyho.ts
gomesalexandre: Nice! Added in 93c3830a36890c851f2ef1c8195b39a517e80a7b

0xApotheosis (approved):

kaladinlight (dismissed):

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -214,6 +215,62 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
     return chainId === 1;
   }
 
+  /*
+   * Tally works the same way as metamask.
+   * This code is copied from the @metamask/detect-provider package
+   * @see https://www.npmjs.com/package/@metamask/detect-provider
+   */
+  private async detectTallyProvider(): Promise<TallyHoEthereumProvider | null> {
```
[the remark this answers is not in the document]
kaladinlight: Sounds like we need the correct metamask package published to leverage any of these ideas so resolving for now until possible. Thanks!

kaladinlight (approved):

0xApotheosis (approved):

PR #541, opened 2022-06-11
feat: add ethSwitchChain method

gomesalexandre:
### Description

This adds a `ethSwitchChain` method to allow switching the currently active wallet chain, using the `wallet_switchEthereumChain` JSON RPC method.

### Issue

Tackles https://github.com/shapeshift/web/issues/1843

#### MetaMask

image

### XDEFI

XDEFI doesn't need user confirmation to switch chains. This prepares `hdwallet-xdefi` for chain switch, but shouldn't be consumed in web since this isn't in the milestone 0.

### TallyHo

~~TallyHo doesn't support multi-chain currently but it is in the roadmap, see the high-level roadmap https://tally-ho.upvoty.com/.~~ 
This prepares `hdwallet-tallyho` for chain switch, but shouldn't be consumed in web currently since calling it will be a no-op and it's not in the milestone 0.
Update: Tally does support multi-chain, but not Avalanche currently, and not in the current stable release. There are [flags](https://github.com/tallycash/extension/blob/cdbeb8b5840caa6fb7f7438d8dc38f681b0dfebf/background/features.ts#L10) to enable Polygon/Arbitrum/Optimism.
Building from source and enabling the right flags gives us this network switcher, so it looks like we might be able to support multiple EVM chains with Tally in the future
image

### DIFF
--- a/packages/hdwallet-core/src/ethereum.ts
+++ b/packages/hdwallet-core/src/ethereum.ts
@@ -108,6 +108,12 @@ export interface ETHWalletInfo extends HDWalletInfo {
   ethSupportsNetwork(chain_id: number): Promise<boolean>;
 
   /**
+   * Switch the wallet's active Ethereum chain
+   * https://eips.ethereum.org/EIPS/eip-3326
+   */
+  ethSwitchChain(chain_id: number): Promise<void>;
+
+  /**
    * Does the device support internal transfers without the user needing to
    * confirm the destination address?
    */
--- a/packages/hdwallet-core/src/ethereum.ts
+++ b/packages/hdwallet-core/src/ethereum.ts
@@ -111,7 +111,7 @@ export interface ETHWalletInfo extends HDWalletInfo {
    * Switch the wallet's active Ethereum chain
    * https://eips.ethereum.org/EIPS/eip-3326
    */
-  ethSwitchChain(chain_id: number): Promise<void>;
+  ethSwitchChain?(chain_id: number): Promise<void>;
 
   /**
    * Does the device support internal transfers without the user needing to
--- a/packages/hdwallet-core/src/ethereum.ts
+++ b/packages/hdwallet-core/src/ethereum.ts
@@ -1,6 +1,13 @@
 import { addressNListToBIP32, slip44ByCoin } from "./utils";
 import { BIP32Path, ExchangeType, HDWallet, HDWalletInfo, PathDescription } from "./wallet";
 
+// https://github.com/MetaMask/eth-rpc-errors/blob/f917c2cfee9e6117a88be4178f2a877aff3acabe/src/classes.ts#L3-L7
+export interface SerializedEthereumRpcError {
+  code: number;
+  message: string;
+  stack?: string;
+}
+
 export enum ETHTransactionType {
   ETH_TX_TYPE_LEGACY = 0,
   ETH_TX_TYPE_EIP_2930 = 1,
--- a/packages/hdwallet-metamask/src/metamask.ts
+++ b/packages/hdwallet-metamask/src/metamask.ts
@@ -1,8 +1,16 @@
+import detectEthereumProvider from "@metamask/detect-provider";
 import * as core from "@shapeshiftoss/hdwallet-core";
+import * as ethers from "ethers";
 import _ from "lodash";
 
 import * as eth from "./ethereum";
 
+// https://github.com/MetaMask/eth-rpc-errors/blob/f917c2cfee9e6117a88be4178f2a877aff3acabe/src/classes.ts#L3-L7
+interface SerializedEthereumRpcError {
+  code: number;
+  message: string;
+  stack?: string;
+}
 export function isMetaMask(wallet: core.HDWallet): wallet is MetaMaskHDWallet {
   return _.isObject(wallet) && (wallet as any)._isMetaMask;
 }
@@ -70,6 +78,21 @@ export class MetaMaskHDWalletInfo implements core.HDWalletInfo, core.ETHWalletIn
 
   public async ethSupportsNetwork(chainId = 1): Promise<boolean> {
     return chainId === 1;
+  }
+
+  public async ethSwitchChain(chainId = 1): Promise<void> {
+    const hexChainId = ethers.utils.hexValue(chainId);
+    // at this point, we know that we're in the context of a valid MetaMask provider
+    try {
+      const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
+      await provider.request({ method: "wallet_switchEthereumChain", params: [{ chainId: hexChainId }] });
+    } catch (e: any) {
+      const error: SerializedEthereumRpcError = e;
+      if (error.code === 4902) {
+        // TODO: EVM Chains Milestone
+        // We will need to pass chainName and rpcUrls, which we don't have yet, to add a chain to MetaMask.
+      }
+    }
   }
 
   public async ethSupportsSecureTransfer(): Promise<boolean> {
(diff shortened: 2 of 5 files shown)


packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -1,8 +1,16 @@
+import detectEthereumProvider from "@metamask/detect-provider";
 import * as core from "@shapeshiftoss/hdwallet-core";
+import * as ethers from "ethers";
 import _ from "lodash";
 
 import * as eth from "./ethereum";
 
+// https://github.com/MetaMask/eth-rpc-errors/blob/f917c2cfee9e6117a88be4178f2a877aff3acabe/src/classes.ts#L3-L7
+interface SerializedEthereumRpcError {
```
gomesalexandre, line 9: Self-review: We probably want a single export for this, to keep things DRY.

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -222,6 +280,59 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
...
+    try {
+      // at this point, we know that we're in the context of a valid MetaMask provider
+      const provider: any = await this.detectTallyProvider();
+      await provider.request({ method: "wallet_switchEthereumChain", params: [{ chainId: hexChainId }] });
+    } catch (e: any) {
+      const error: core.SerializedEthereumRpcError = e;
+      if (error.code === 4902) {
+        // TODO: EVM Chains Milestone
+        // We will need to pass chainName and rpcUrls, which we don't have yet, to add a chain to Tally.
+      }
```
pastaghost, line 332: Log the error for both the 4902-specific and all other cases to ease debugging.

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -63,6 +65,21 @@ export class XDEFIHDWalletInfo implements core.HDWalletInfo, core.ETHWalletInfo
...
+      await provider.request({ method: "wallet_switchEthereumChain", params: [{ chainId: hexChainId }] });
+    } catch (e: any) {
+      const error: core.SerializedEthereumRpcError = e;
+      if (error.code === 4902) {
+        // TODO: EVM Chains Milestone
+        // We will need to pass chainName and rpcUrls, which we don't have yet, to add a chain to XDEFI.
+      }
+    }
+  }
+
```
pastaghost, line 82: I'm not sure we need this method added to the XDEFIHDWalletInfo class.

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -220,6 +237,21 @@ export class XDEFIHDWallet implements core.HDWallet, core.ETHWallet {
...
```
pastaghost, line 254: Log the error for both the 4902-specific and all other cases to ease debugging.

pastaghost (changes requested):

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -62,6 +64,62 @@ export class TallyHoHDWalletInfo implements core.HDWalletInfo, core.ETHWalletInf
...
+    }
+  }
+
```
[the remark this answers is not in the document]
gomesalexandre: Seems like we don't need it there, the call worked without. Removed in de254daa059182713ac2e6c2d17869ff1fc55840

packages/hdwallet-xdefi/src/xdefi.ts
gomesalexandre: Added a console.error in de254daa059182713ac2e6c2d17869ff1fc55840

packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -72,6 +74,21 @@ export class MetaMaskHDWalletInfo implements core.HDWalletInfo, core.ETHWalletIn
...
+      await provider.request({ method: "wallet_switchEthereumChain", params: [{ chainId: hexChainId }] });
+    } catch (e: any) {
+      const error: core.SerializedEthereumRpcError = e;
+      if (error.code === 4902) {
+        // TODO: EVM Chains Milestone
+        // We will need to pass chainName and rpcUrls, which we don't have yet, to add a chain to MetaMask.
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
gomesalexandre: Seems like we don't need it there, the call worked without. Removed in de254daa059182713ac2e6c2d17869ff1fc55840

packages/hdwallet-xdefi/src/xdefi.ts
gomesalexandre: Seems like we don't need it there, the call worked without. Removed in de254daa059182713ac2e6c2d17869ff1fc55840

packages/hdwallet-tallyho/src/tallyho.ts
gomesalexandre: Added a console.error in https://github.com/shapeshift/hdwallet/commit/de254daa059182713ac2e6c2d17869ff1fc55840.

pastaghost (approved):

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -222,6 +224,61 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
     return false;
   }
 
+  private async detectTallyProvider(): Promise<TallyHoEthereumProvider | null> {
```
0xApotheosis, line 227: Add the below comment?

```ts
  /*
   * Tally works the same way as metamask.
   * This code is copied from the @metamask/detect-provider package
   * @see https://www.npmjs.com/package/@metamask/detect-provider
   */
```

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -1,6 +1,13 @@
 import { addressNListToBIP32, slip44ByCoin } from "./utils";
 import { BIP32Path, ExchangeType, HDWallet, HDWalletInfo, PathDescription } from "./wallet";
 
+// https://github.com/MetaMask/eth-rpc-errors/blob/f917c2cfee9e6117a88be4178f2a877aff3acabe/src/classes.ts#L3-L7
+export interface SerializedEthereumRpcError {
```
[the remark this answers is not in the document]
gomesalexandre: This isn't intended to be some 1/1 typing, but a type that works for us. But reading back into this, I can see the value of having the full type

packages/hdwallet-tallyho/src/tallyho.ts
gomesalexandre: Added in 2f22d667a2298e09b5215abb41de74343a6b0ce1

packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -249,6 +251,22 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
     return chainId === 1;
   }
 
+  public async ethSwitchChain(chainId: number): Promise<void> {
+    const hexChainId = ethers.utils.hexValue(chainId);
+    const provider: any = await detectEthereumProvider({ mustBeMetaMask: true, silent: false, timeout: 3000 });
```
kaladinlight, line 256: might be useful to create an ethereum provider interface that we can pass in as the generic return type that types everything from the metamask provider (https://github.com/MetaMask/detect-provider/blob/5ce916fc24779c4b36741531a41d9c8b3cbb0a17/src/index.ts#L1) and adds the request func typings: https://docs.metamask.io/guide/ethereum-provider.html#ethereum-request-args

kaladinlight: Or looks like that ethers.externalProvider would be even better

packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -249,6 +251,22 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
...
+    try {
+      // at this point, we know that we're in the context of a valid MetaMask provider
+      await provider.request({ method: "wallet_switchEthereumChain", params: [{ chainId: hexChainId }] });
+    } catch (e: any) {
```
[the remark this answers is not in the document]
gomesalexandre: Since adding a chain if not added in MetaMask yet is a new feature (which looks like error handling here), I'll use MetaMask's type guard in the follow-up PR that will add the (C- in our case) chain if missing

kaladinlight (approved):

kaladinlight (approved):

0xApotheosis (approved):
🍭

PR #546, opened 2022-06-29
feat: add bytes field to ethereum signMessage methods

chxash:
This adds the bytes possibility for messages in `ETHSignMessage` input of `ethSignMessage` methods.
Basically there is no change related to current behaviour with `message` as a string.
But now we are able to pass message as `Bytes` (= `ArrayLike<number>`).

This is required to sign messages for cowswap at least because of the following :
"A common case is to sign a hash. In this case, if the hash is a string, it must be converted to an array first, using the [arrayify](https://docs.ethers.io/v5/api/utils/bytes/#utils-arrayify) utility function."
(from https://docs.ethers.io/v5/api/signer/#Signer-signMessage)

I linked to a custom web in order to test some signatures.
I tested signatures generated from `bytes` input for `Native` and `Metamask` and I was able to get signatures valid for cowswap.
I was not able to get valid signature for Metamask + ledger so probably more work will be needed for this (eip712 implementation ?)

Needed for https://github.com/shapeshift/lib/issues/726

**Testing steps:**

- Build following lib branch with linked hdwallet branch https://github.com/shapeshift/lib/tree/cowswap-executeTrade

- Build following web branch with linked lib / hdwallet https://github.com/shapeshift/web/tree/cowswapper-web-integration-wip . This is a web branch where CowSwapper is used as primary swapper (so it will be used in priority for any erc20 swaps), if you try to trade an ETH pair 0x will be used.
I used local linking syntax `"@shapeshiftoss/hdwallet-core": "../hdwallet/packages/hdwallet-core"` but feel free to do something else.

- Launch web branch locally with native wallet loaded with some ETH / erc20 token (ex : WETH)

- In trade component : trade for example WETH / FOX (erc20 are required to use cowswap) and click on preview trade

- Then click on confirm trade (I have put a glitch in lib so the orders wont be accepted by cowswap , you can have the glitch removed by changing line 46 in `lib/CowExecuteTrade.ts` `validTo: 1656667297, //getNowPlusThirtyMinutesTimestamp()`, use the commented line instead of 1656667297 . Be careful that polling for order completion is not implemented yet on that lib branch).

- If signature is considered valid by cowswap, you should notice a 400 error in network tab for `/orders` api call with following error : `{"errorType":"InsufficientValidTo","description":"validTo is not far enough in the future"}`
If signature is deemed invalid by cowswap, you should receive the following error : `{"errorType":"WrongOwner","description":"Address recovered from signature 0x4xxx…yyy does not match from address"}`

https://api.cow.fi/docs/#/default/post_api_v1_orders can be used to see the various results of the api and verify that by changing the signature we effectively get the mentionned errors.


packages/hdwallet-ledger/src/ethereum.ts
```diff
@@ -150,7 +150,9 @@ export async function ethSignMessage(
   msg: core.ETHSignMessage
 ): Promise<core.ETHSignedMessage> {
   const bip32path = core.addressNListToBIP32(msg.addressNList);
-  const res = await transport.call("Eth", "signPersonalMessage", bip32path, Buffer.from(msg.message).toString("hex"));
+
+  const buffer = typeof msg.message === "string" ? Buffer.from(msg.message) : Buffer.from(new Uint8Array(msg.message));
```
gomesalexandre, line 154: If possible, we should use `ethers.utils.arrayify`/ `core.arrayify` instead of `Uint8Array`, which effectively does the same but is a battle-tested, nice abstraction to pack things

gomesalexandre: With that being said, not sure if this really matters in this module, `@shapeshiftoss/hdwallet-ledger` isn't used by our projects, and I'm not sure it's maintained. @pastaghost might have some context on whether or not it is (supposed to be) used by the community

gomesalexandre (commented):
As we've paired on this one today, that looks good from a high-level of passing `Bytes` instead of an hex string, as their `sizeof` is different and the hex string representation won't produce a valid `signMessage()` signature. At worst, this won't regress the current implementation so whether or not this consistently works across wallets or is EIP-712 compliant, happy with it as a base to progress on Cowswap signing. 


Could you add testing steps to test this one out (e.g from web)? 🙏🏽

packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -189,7 +189,7 @@ export async function ethGetAddress(transport: Transport, msg: core.ETHGetAddres
 export async function ethSignMessage(transport: Transport, msg: core.ETHSignMessage): Promise<core.ETHSignedMessage> {
   const m = new Messages.EthereumSignMessage();
   m.setAddressNList(msg.addressNList);
-  m.setMessage(toUTF8Array(msg.message));
+  m.setMessage(typeof msg.message === "string" ? toUTF8Array(msg.message) : new Uint8Array(msg.message));
```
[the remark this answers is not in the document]
chxash: Tried it but there is a compilation error since setMessage is expecting `string | Uint8Array`

packages/hdwallet-ledger/src/ethereum.ts
chxash: @gomesalexandre I don't think we can use `ethers.utils.arrayify/ core.arrayify` here since msg.message is already of types `Bytes`. Those methods are expecting a string as input. 
We are simply transforming the `Bytes` input to an `Uint8Array` which is pretty straightforward I think.

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -83,9 +83,11 @@ export interface ETHSignedTx {
   serialized: string;
 }
 
+export type Bytes = ArrayLike<number>;
```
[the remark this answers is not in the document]
chxash: done !

chxash:
@gomesalexandre updated the PR description with testing steps ! feel free to ping me for any help

gomesalexandre (dismissed):
LGTM, saw usage during a call with https://github.com/shapeshift/lib/pull/868 linked, and a web branch consuming both.

Already reviewed this previously and confirmed this is a working implementation to sign messages, that is not producing any regressions with the existing implementations (furthermore, we're not using `ethSignMessage`/`signMessage` anywhere currently, although *that alone would not be a criteria for accepting this, since we have many other consumers as hdwallet as an open-source project*)

Would rather use `isBytes` than checking for `typeof x === 'string'` to be more explicit, but overall looks good! 🌮

gomesalexandre (dismissed):

0xdef1cafe:
@chxash this looks conceptually fine, and i'll otherwise approve, two small things

* a hdwallet version was published yesterday at `1.26.0` - you'll need to bump again to `1.27.0`
* you're failing build in CI currently

chxash:
@0xdef1cafe @gomesalexandre  After merging latest master I had a build error related to latest merged PR.
I fixed the issue with latest commit : https://github.com/shapeshift/hdwallet/pull/546/commits/00cc62c6d540b40877eb682ab922e986e61aa279
I applied very similar logic to what I have been doing elsewhere.
@pastaghost is it fine for you ?

0xdef1cafe (approved):

PR #547, opened 2022-06-29
feat: add ethAddChain method

gomesalexandre:
This adds an `ethAddChain` that allows adding a new Ethereum chain, and consumes it in MetaMask's `eth_switchChain` in case the chain we're switching to isn't currently added.

Note that this is only consumed in MetaMask: while switching chains is a valid operation in XDEFI and Tally, adding a network in unsupported in Both (although Tally explicitly has it as an unsupported method for now, hinting at future support https://github.com/tallycash/extension/blob/a7a7c1148025e27f522db201e49a0cb711fef9cd/background/services/internal-ethereum-provider/index.ts#L277)

image

### DIFF
--- a/packages/hdwallet-core/src/ethereum.ts
+++ b/packages/hdwallet-core/src/ethereum.ts
@@ -106,6 +106,20 @@ export interface ETHVerifyMessage {
   signature: string;
 }
 
+// https://docs.metamask.io/guide/rpc-api.html#wallet-addethereumchain
+export interface AddEthereumChainParameter {
+  chainId: string; // A 0x-prefixed hexadecimal string
+  chainName: string;
+  nativeCurrency: {
+    name: string;
+    symbol: string; // 2-6 characters long
+    decimals: 18;
+  };
+  rpcUrls: string[];
+  blockExplorerUrls?: string[];
+  iconUrls?: string[]; // Currently ignored.
+}
+
 export interface ETHWalletInfo extends HDWalletInfo {
   readonly _supportsETHInfo: boolean;
 
@@ -125,6 +139,12 @@ export interface ETHWalletInfo extends HDWalletInfo {
    * https://eips.ethereum.org/EIPS/eip-3326
    */
   ethSwitchChain?(chain_id: number): Promise<void>;
+
+  /**
+   * Add an Ethereum chain to user's wallet
+   * https://eips.ethereum.org/EIPS/eip-3085
+   * */
+  ethAddChain?(params: AddEthereumChainParameter): Promise<void>;
 
   /**
    * Does the device support internal transfers without the user needing to
--- a/packages/hdwallet-metamask/src/metamask.ts
+++ b/packages/hdwallet-metamask/src/metamask.ts
@@ -1,8 +1,23 @@
 import * as core from "@shapeshiftoss/hdwallet-core";
+import { AddEthereumChainParameter } from "@shapeshiftoss/hdwallet-core";
+import { serializeError } from "eth-rpc-errors";
 import * as ethers from "ethers";
 import _ from "lodash";
 
 import * as eth from "./ethereum";
+
+// https://docs.avax.network/dapps/smart-contracts/add-avalanche-to-metamask-programmatically
+const AVALANCHE_MAINNET_ADD_CHAIN_PARAMS: AddEthereumChainParameter = {
+  chainId: "0xA86A",
+  chainName: "Avalanche Mainnet C-Chain",
+  nativeCurrency: {
+    name: "Avalanche",
+    symbol: "AVAX",
+    decimals: 18,
+  },
+  rpcUrls: ["https://api.avax.network/ext/bc/C/rpc"],
+  blockExplorerUrls: ["https://snowtrace.io/"],
+};
 
 export function isMetaMask(wallet: core.HDWallet): wallet is MetaMaskHDWallet {
   return _.isObject(wallet) && (wallet as any)._isMetaMask;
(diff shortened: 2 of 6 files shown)


packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -261,18 +276,30 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
...
-        // We will need to pass chainName and rpcUrls, which we don't have yet, to add a chain to MetaMask.
+      const error = serializeError(e);
+      if (error.code === -32603) {
+        // We only support Avalanche C-Chain currently. It is supported natively in XDEFI, and unsupported in Tally, both with no capabilities to add a new chain
+        // TODO(gomes): Find a better home for these. When that's done, we'll want to call ethSwitchChain with (params: AddEthereumChainParameter) instead
+        await this.ethAddChain(AVALANCHE_MAINNET_ADD_CHAIN_PARAMS);
+        return;
       }
+
+      throw (error.data as any).originalError as {
```
gomesalexandre, line 298: `serializeError` improperly types `data` as unknown, see https://github.com/MetaMask/eth-rpc-errors/blob/f917c2cfee9e6117a88be4178f2a877aff3acabe/src/classes.ts#L6

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -126,6 +140,12 @@ export interface ETHWalletInfo extends HDWalletInfo {
    */
   ethSwitchChain?(chain_id: number): Promise<void>;
 
+  /**
+   * Add an Ethereum chain to user's wallet
+   * https://eips.ethereum.org/EIPS/eip-3085
+   * */
+  ethAddChain?(params: AddEthereumChainParameter): Promise<void>;
```
0xApotheosis, line 147: Should this be `evmAddChain` or `ethBasedAddChain` instead?

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -106,6 +106,20 @@ export interface ETHVerifyMessage {
   signature: string;
 }
 
+// https://docs.metamask.io/guide/rpc-api.html#wallet-addethereumchain
+export interface AddEthereumChainParameter {
```
0xApotheosis, line 110: Likewise, `AddEthereumBasedChainParameter` or `AddEvmChainParameter`?

cjthompson:
See my [other comment](https://github.com/shapeshift/hdwallet/pull/548#issuecomment-1180874881)

packages/hdwallet-core/src/ethereum.ts
cjthompson: > Should this be `evmAddChain` or `ethBasedAddChain` instead?

I agree with you in principle but it would break the established pattern of all the functions being named `ethXXX`. For now, I don't think it's worth introducing an inconsistency.

gomesalexandre:
> See my [other comment](https://github.com/shapeshift/hdwallet/pull/548#issuecomment-1180874881)

@cjthompson Merged the diff from https://github.com/shapeshift/hdwallet/pull/548 (including package bump/release that will now happen on this PR instead) and renamed `supportsAvalanche` occurrences to `supportsEthSwitchChain`, including renaming the `supportsEthSwitchChain` type guard: 2278492

packages/hdwallet-core/src/wallet.ts
```diff
@@ -137,6 +137,10 @@ export function supportsCosmos(wallet: HDWallet): wallet is CosmosWallet {
   return _.isObject(wallet) && (wallet as any)._supportsCosmos;
 }
 
+export function supportsEthSwitchChain(wallet: HDWallet): wallet is ETHWallet {
```
0xApotheosis, line 140: Something feels off about having this as a type guard - it isn't actually a type guard for `ETHWallet`, it's more of a helper with a `boolean` return type.

E.g. if `supportsEthSwitchChain` returns false it does not mean the `HDWallet` object it received is _not_ an `ETHWallet`, only that it does not support ETH switch chain.

That said, I actually don't think any of these functions are _actually_ type guards in any way except for their syntax - but rather mere helper functions.

Probably best to just continue propagating the existing pattern for now...

packages/hdwallet-metamask/src/metamask.ts
```diff
@@ -261,18 +277,30 @@ export class MetaMaskHDWallet implements core.HDWallet, core.ETHWallet {
...
       // at this point, we know that we're in the context of a valid MetaMask provider
       await this.provider.request({ method: "wallet_switchEthereumChain", params: [{ chainId: hexChainId }] });
     } catch (e: any) {
-      const error: core.SerializedEthereumRpcError = e;
-      console.error(error);
-      if (error.code === 4902) {
-        // TODO: EVM Chains Milestone
-        // We will need to pass chainName and rpcUrls, which we don't have yet, to add a chain to MetaMask.
+      const error = serializeError(e);
+      if (error.code === -32603) {
```
0xApotheosis, line 292: Same thought here about the magic number error code.

0xApotheosis (approved):

packages/hdwallet-core/src/wallet.ts
gomesalexandre: Agreed these should not be type guards to start with. It's not *wrong* per se (if this method returns true, it is an `EthWallet`), it just makes no sense for this to be a type guard, as this discriminates on something entirely different.

You're right consistency > cleanness, we will need to do grep through all the type guards and remove the ones where the guard doesn't actually guard against anything

packages/hdwallet-metamask/src/metamask.ts
gomesalexandre: Added in eb3093a225d13b9e439f28cf3addec817661cbd8

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -242,10 +243,11 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
     } catch (e: any) {
       const error: core.SerializedEthereumRpcError = e;
       console.error(error);
-      if (error.code === 4902) {
-        // TODO: EVM Chains Milestone
-        // We will need to pass chainName and rpcUrls, which we don't have yet, to add a chain to Tally.
+      if (error.code === -32603) {
```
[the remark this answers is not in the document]
gomesalexandre: Added in eb3093a225d13b9e439f28cf3addec817661cbd8

packages/hdwallet-keepkey/src/keepkey.ts
```diff
@@ -623,6 +623,7 @@ export class KeepKeyHDWallet implements core.HDWallet, core.BTCWallet, core.ETHW
   readonly _supportsDebugLink: boolean;
   readonly _isKeepKey = true;
   readonly _supportsETH = true;
+  readonly _supportsEthSwitchChain = true;
```
0xdef1cafe, line 626: is this really true for keepkey? @pastaghost @BitHighlander think this should be false right now

packages/hdwallet-native/src/native.ts
```diff
@@ -210,6 +210,7 @@ export class NativeHDWallet
   readonly _supportsBTC = true;
   readonly _supportsETH = true;
   readonly _supportsCosmos = true;
+  readonly _supportsEthSwitchChain = true;
```
0xdef1cafe, line 213: i don't think this is true yet until we impl avalanche support

packages/hdwallet-native/src/ethereum.ts
```diff
@@ -52,6 +52,7 @@ export function MixinNativeETHWallet<TBase extends core.Constructor<NativeHDWall
   // eslint-disable-next-line @typescript-eslint/no-shadow
   return class MixinNativeETHWallet extends Base {
     readonly _supportsETH = true;
+    readonly _supportsEthSwitchChain = true;
```
0xdef1cafe, line 55: remove for now until this is true

packages/hdwallet-native/src/native.ts
gomesalexandre: Removed in b3852ba8a44e80c11c1a5f2804dbcf687d3b4116

gomesalexandre: It will probably never be true either way as we don't switch chains per se with native

packages/hdwallet-native/src/ethereum.ts
gomesalexandre: Removed in b3852ba8a44e80c11c1a5f2804dbcf687d3b4116

0xdef1cafe (approved):

0xApotheosis (approved):
Thanks for the updates 🙏  LGTM

PR #548, opened 2022-06-29
feat: add wallet.supportsAvalanche() method

gomesalexandre:
### Description

This adds `_supportsAvalanche` / `_supportsAvalancheInfo` class members and consumes the former in the `wallet.supportsAvalanche()` method 

### Issue

- partially tackles https://github.com/shapeshift/web/issues/2097


packages/hdwallet-core/src/ethereum.ts
```diff
@@ -108,6 +108,7 @@ export interface ETHVerifyMessage {
 
 export interface ETHWalletInfo extends HDWalletInfo {
```
0xApotheosis, line 109: Should this be changed to `EvmWalletInfo`?

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -160,6 +161,7 @@ export interface ETHWalletInfo extends HDWalletInfo {
 
 export interface ETHWallet extends ETHWalletInfo, HDWallet {
```
0xApotheosis, line 162: Should this be changed to `EvmWallet`?

cjthompson:
Since there isn't any implementation changes in this, I don't think we need to have a `support` method for every EVM chain out there.  Instead, I think you should combine this with #547 and rename `supportAvalanche` to `supportsEthAddChain`. You could even add `supportsEthSwitchChain` (there is an `ethSwitchChain` method above the new you one you added).

Then you just do feature detection for whether `ethAddChain` and/or `ethSwitchChain` is supported and do the thing accordingly.

Then you can add future EVM chain support to any wallet that implements those methods.

packages/hdwallet-core/src/ethereum.ts
cjthompson: > Should this be changed to `EvmWallet`?

Yes, but it would be a massive breaking change. It would be better to just add a JSDoc comment above explaining that it's really for any Ethereum-compatible chain.

gomesalexandre:
> Then you just do feature detection for whether `ethAddChain` and/or `ethSwitchChain` is supported and do the thing accordingly.

That sounds like a good option to me. While this made sense for native (or base in the context of the EVMBaseAdapter) chains, you're right we shouldn't need to add these for every EVM chains we add.
`supportsEthSwitchChain` will do the trick, and make feature detection easier for https://github.com/shapeshift/web/issues/2130

gomesalexandre:
Closing in favor of `supportsEthSwitchChain` in https://github.com/shapeshift/hdwallet/pull/547

Issue #549, opened 2022-07-12
Remove unused type guards


gomesalexandre:
### Overview

We define wallet methods as type guards in `hdwallet-core`

However, we don't use them as type guards but only as regular utility methods to check for feature detection e.g

https://github.com/shapeshift/hdwallet/blob/9ada7ded938f7a8ee752517523a724d6a18d5f3e/packages/hdwallet-core/src/wallet.ts#L120-L122

We should remove the unused type guard in all these methods.

### References and additional details

hdwallet-core methods implementations:

https://github.com/shapeshift/hdwallet/blob/master/packages/hdwallet-core/src/wallet.ts

### Acceptance Criteria

- The hdwallet-core type guards are removed

0xApotheosis:
Thank you for this @gomesalexandre!

PR #435, opened 2022-02-12
refactor: remove `isWalletType` introspection functions

mrnerdhair:
`instanceof` makes implementing new wallet types easier (and everything less messy)

PR #550, opened 2022-07-13
feat: remove unused type guard typings

gomesalexandre:
## Description

This removes the unused type guard (`x is Y`) from the `supportsXYZ` methods

Note that the `supportsCosmos` and `supportsOsmosis` ones are kept, as they are actually used in `chain-adapters`.

Tried linking in lib and web and this builds/passes type-check.

## Issue

- closes https://github.com/shapeshift/hdwallet/issues/549

### DIFF
--- a/packages/hdwallet-core/src/wallet.ts
+++ b/packages/hdwallet-core/src/wallet.ts
@@ -1,18 +1,8 @@
 import _ from "lodash";
 
-import { BinanceWallet, BinanceWalletInfo } from "./binance";
 import { BTCInputScriptType, BTCWallet, BTCWalletInfo } from "./bitcoin";
-import { CosmosWallet, CosmosWalletInfo } from "./cosmos";
-import { DebugLinkWallet } from "./debuglink";
-import { EosWallet, EosWalletInfo } from "./eos";
-import { ETHWallet, ETHWalletInfo } from "./ethereum";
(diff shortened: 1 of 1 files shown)


gomesalexandre:
@pastaghost Keeping this as a draft waiting for your input here. In the context of our consumptions, the type guards are effectively useless, but now that I'm thinking about it, we might have other projects consuming this and it might bring breaking type changes for them.

Happy to close both this and the issue with the rationale that hdwallet is open-source and we shouldn't tie implementations to our own usages.

0xApotheosis:
@pastaghost thoughts on this one re: @gomesalexandre comment?

I'm in favour of removing the type guard typings, as these functions aren't actually type guards (beyond their syntax), and so are misleading.

0xApotheosis (approved):
I think this PR makes these functions more correct, as they are more utility functions than actual type guards.

Pending thoughts from @pastaghost.

pastaghost (approved):

pastaghost:
I agree with @0xApotheosis on this one - I don't think this is proper use of a user-defined type guard. I'm in favor of removing the type guard returns and have approved the PR.

mrnerdhair:
(For an example of the above, see #435, which was closed at the time due to lack of review bandwidth.)

PR #554, opened 2022-07-26
fix: kk non-segwit utxo tx signing

asamere:
don't double sats convert utxo referenced prevTx amount @BitHighlander


elmutt (dismissed):

gomesalexandre (approved):
Very specific logic bit so untested, but looking at the `satsFromStr` function definition in https://github.com/shapeshift/hdwallet/blob/805617c2018543d611522d69d6b4e063ec9755b5/packages/hdwallet-core/src/utils.ts#L158 and the intent of this PR, this looks sound.

gomesalexandre:
@asamere Given this is only used in one place currently, do we want to remove the `satsFromStr` definition in a follow-up PR? I'm always cautious about such breaking changes, as, as @pastaghost mentioned, many projects are consuming hdwallet packages, not just our own.

With [a global GitHub search for it](https://github.com/search?q=satsFromStr&type=code), I can see that the only references to it is on one fork of hdwallet, which hasn't been kept up to date with origin upstream in months, with the only consumption being the one that this removes so IMO it should be safe to deprecate/remove.

0xdef1cafe (approved):

0xdef1cafe:
nice find! tests are sad now and look related and real failures.

kaladinlight:
> @asamere Given this is only used in one place currently, do we want to remove the `satsFromStr` definition in a follow-up PR? I'm always cautious about such breaking changes, as, as @pastaghost mentioned, many projects are consuming hdwallet packages, not just our own.
> 
> With [a global GitHub search for it](https://github.com/search?q=satsFromStr&type=code), I can see that the only references to it is on one fork of hdwallet, which hasn't been kept up to date with origin upstream in months, with the only consumption being the one that this removes so IMO it should be safe to deprecate/remove.

In addition to this, considering this logic has been in place since HDWallet was created, I would be thinking along the same lines of @gomesalexandre unless we have not been able to send legacy since the inception of HDWallet. Does seem weird for a wallet api to want to talk in terms of btc over sats though regardless

kaladinlight (approved):

asamere:
> nice find! tests are sad now and look related and real failures.

taking a look at tests

elmutt (dismissed):

kaladinlight:
> > 
> 
> @kaladinlight @gomesalexandre did CQ by chance return prevTx value as floating point? this toSats/fromSats nonsense is a relic of hurried axiom dev from my recollection.

cq just talking in sats, unless watchtower was doing anything weird. was legacy send broken forever? that is what I am a little confused about

kaladinlight (dismissed):

kaladinlight (approved):

kaladinlight:
We can remove the `satsFromStr` helper as well as it is no longer in use?

0xdef1cafe (approved):

PR #556, opened 2022-08-07
fix: wallet integration tests

gomesalexandre:
This fixes the integration tests not passing on the last builds, see https://github.com/shapeshift/hdwallet/runs/7708987352


0xApotheosis (commented):
I tried to run this locally, and it seems we still need to update all of the `integration/src/{wallet}/index.ts` files, too?

That said, this it could be a local env thing because CI passes.

Screen Shot 2022-08-08 at 6 25 07 am

pastaghost:
@gomesalexandre I can also confirm @0xApotheosis' result - I see the same errors when running `yarn test:integration` locally.

0xApotheosis:
@pastaghost weird how CI passes though, is that expected?

pastaghost:
> @pastaghost weird how CI passes though, is that expected?

@0xApotheosis I think CircleCI is misconfigured for non-master branches in this repo. Both the `build` and `release-packages` are run on the master branch, but  looking at the last CI run on the `fix_wallet_integration_tests` branch, it looks like neither of the CircleCI jobs was executed. That would certainly explain why @gomesalexandre's branch seems to be passing CI.

gomesalexandre:
> I tried to run this locally, and it seems we still need to update all of the `integration/src/{wallet}/index.ts` files, too?
> 
> That said, this it could be a local env thing because CI passes.

I just tried deleting and re-cloning the repo as a sanity check and running `yarn test:integration` off freshly built packages in master which gave me this, and the exact same in this branch:

```sh
Test Suites: 1 failed, 6 passed, 7 total
Tests:       94 failed, 54 skipped, 515 passed, 663 total
```

But you're right these need to be updated too. I assume it all to be some hdwallet build shenanigans, since actually going to `integration/src/{wallet}/index.ts` files triggers my languageserver to report the errors you've mentioned:

image

Tackled in 26a19af4d0ebfe031d07a8c0acff7cbc60919ab9 - @0xApotheosis and @pastaghost could you give it another run locally in case I can't fix the local shenanigans by then? 🙏🏽

Alternatively, is there a way to trigger CI on non-master branches @pastaghost?

0xApotheosis:
Hmm, now the language server is happy, but `yarn test:integration` still is still sad (small extract. below):

Screen Shot 2022-08-08 at 7 58 04 pm

0xApotheosis (approved):
I do not know if this will fix CI on `master`, but it's strictly better than `master` at the moment.

I'm still getting failing tests when running `yarn test:integration`, but I also get that on `master`, so it's not a regression in this PR.

0xApotheosis:
> I think CircleCI is misconfigured for non-master branches in this repo.

Thanks for the context @pastaghost. I just took a look at CircleCI to see if I could fix the issue, but it seems I don't have the needed permissions to access the `hdwallet` pipeline.

Assuming you do have the needed permissions, is this something you can take a look at?
Want me to make a ticket for it?

PR #568, opened 2022-09-07
feat: add wallet supports multi account function

### LINKED ISSUE #567: Add concept of "wallet supports multi account"
### Overview

Some new(er) wallets we support post-axiom don't support the concept of multiple accounts, e.g. MetaMask-like wallets - Tally Ho, WalletConnect, Keplr etc

### References and additional details

We need a way to query if the currently connected wallet supports multiple accounts, or only a single account at the same time.

### Acceptance Criteria

* We can ask the wallet if it supports multiple chains

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

asamere:
adds a new function `supportsBip44Accounts(): boolean` to the HDWallet interface via HDWalletInfo and implements for all supported wallets.

usage from web:
```
if (wallet.supportsBIP44Accounts()) {
  // hoorah do multi account things
}
```      

closes #567


pastaghost (dismissed):

packages/hdwallet-portis/src/portis.ts
```diff
@@ -45,6 +45,10 @@ export class PortisHDWalletInfo implements core.HDWalletInfo, core.ETHWalletInfo
     return false;
   }
 
+  public supportsBip44Accounts(): boolean {
+    return true; // I think?
```
0xdef1cafe, line 49: false i think

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -33,10 +33,14 @@ export class TallyHoHDWalletInfo implements core.HDWalletInfo, core.ETHWalletInf
   }
 
   public hasNativeShapeShift(): boolean {
-    // It doesn't... yet?
+    // It doesn't... yet? - never ever never ever ever
```
0xdef1cafe, line 36: correct, false, it's metamask-like

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -37,6 +37,10 @@ export class XDEFIHDWalletInfo implements core.HDWalletInfo, core.ETHWalletInfo
     return false;
   }
 
+  public supportsBip44Accounts(): boolean {
+    return true; // probably?
```
0xdef1cafe, line 41: requires testing

packages/hdwallet-keplr/src/keplr.ts
```diff
@@ -55,6 +55,10 @@ export class KeplrHDWalletInfo implements core.HDWalletInfo, core.CosmosWalletIn
     return false;
   }
 
+  public supportsBip44Accounts(): boolean {
+    return true;
```
0xdef1cafe, line 59: def false for this one - also metamask-like

0xdef1cafe (changes requested):

packages/hdwallet-tallyho/src/tallyho.ts
asamere: "NativeShapeShift" is (was) a feature where KK can show the details of an old-school ShapeShift "shift" similar to the way it shows details of an erc20 send for ETH but you just get blind signing for other ETH stuff. There was a whole extra protobuf signed by and served from our backend the KK would verify signature of and show.

packages/hdwallet-portis/src/portis.ts
asamere: damn it you're probably right. I'll test

packages/hdwallet-xdefi/src/xdefi.ts
0xdef1cafe: this needs to be false - it supports multiple accounts but not at the same time.

only native + keepkey support them at the same time currently

asamere: 🙏

packages/hdwallet-portis/src/portis.ts
asamere, line 49: portis not loading in beta or v2 to try this out at the moment

integration/src/wallets/metamask.ts
```diff
@@ -66,6 +66,11 @@ export function selfTest(get: () => core.HDWallet): void {
     expect(wallet.ethSupportsNativeShapeShift()).toEqual(false);
   });
 
+  it("does not supports bip44 accounts", async () => {
+    if (!wallet) return;
+    expect(wallet.supportsBip44Accounts()).toEqual(false);
```
pastaghost, line 71: This is a tricky one - when the snap is released, we will be able to provide BIP44 account support through MetaMask for all chains supported by the snap (this includes ETH.) 

We should probably add a conditional return here. If the MetaMask provider has the ShapeShift snap installed and activated, return true. Otherwise, return false. I provide a helper function in the ShapeShift snap adapter that does exactly this.

integration/src/wallets/native.ts
```diff
@@ -342,6 +347,10 @@ export function selfTest(get: () => core.HDWallet): void {
     ]);
   });
 
+  it("returns true for supportsBip44Accounts", () => {
```
kaladinlight, line 350: redundant test?

kaladinlight (approved):
why the `Wallet` does not extend `WalletInfo` I don't know, and lot of semi silly tests, but looks reasonable to me

0xdef1cafe (approved):

kaladinlight (dismissed):

0xdef1cafe (changes requested):
package bump

integration/src/wallets/native.ts
asamere: this the info vs regular thing?

0xdef1cafe (approved):

PR #571, opened 2022-09-13
chore: enable BIP44 Accounts other than zero for Ethereum in Native Wallet

asamere:
Enables Bip44 Multi Account support for Ethereum (and evm chains) in HDWallet Native.


packages/hdwallet-native/src/ethereum.ts
```diff
@@ -57,7 +57,10 @@ export function MixinNativeETHWallet<TBase extends core.Constructor<NativeHDWall
 
     async ethInitializeWallet(masterKey: Isolation.Core.BIP32.Node): Promise<void> {
       const rootNode = await Isolation.Adapters.BIP32.create(masterKey);
-      const isolatedSigner = await rootNode.derivePath(ethers.utils.defaultPath);
+
+      // hard coding this path works fine. need to figure out how to
+      // re-enable supplying a path
```
asamere, line 62: we need to derive the rootNode from a depth 5 path with the account number at index 2. this is due to upstream requirements of ethers.js.

0xdef1cafe (approved):

packages/hdwallet-native/src/ethereum.ts
```diff
@@ -53,32 +55,43 @@ export function MixinNativeETHWallet<TBase extends core.Constructor<NativeHDWall
...
 
     async ethGetAddress(msg: core.ETHGetAddress): Promise<string | null> {
-      if (!_.isEqual(msg.addressNList, core.bip32ToAddressNList("m/44'/60'/0'/0/0"))) {
-        throw new Error("path not supported");
+      if (msg.addressNList.length < 5) {
+        throw new Error(`path not supported length: ${msg.addressNList.length}`);
+      }
+
+      // regex sux
+      const refAddressNList = core.bip32ToAddressNList("m/44'/60'/0'/0/0");
```
asamere, line 76: need to make this support chains besides ETH

0xdef1cafe (changes requested):
disgusting

asamere:
> 

🤮

packages/hdwallet-native/src/crypto/isolation/adapters/ethereum.ts
```diff
@@ -9,18 +10,20 @@ function ethSigFromRecoverableSig(x: SecP256K1.RecoverableSignature): ethers.Sig
...
-    super();
-    this._isolatedKey = isolatedKey;
+  protected constructor(rootNode: Isolation.Adapters.BIP32, provider?: ethers.providers.Provider) {
+    this.rootNode = rootNode;
     this.provider = provider;
   }
 
-  static async create(isolatedKey: SecP256K1.ECDSAKey, provider?: ethers.providers.Provider): Promise<SignerAdapter> {
-    return new SignerAdapter(isolatedKey, provider);
+  static async create(
```
kaladinlight, line 22: we don't really need the static async create anymore if we want to clean that up

packages/hdwallet-native/src/crypto/isolation/adapters/ethereum.ts
```diff
@@ -32,13 +35,15 @@ export class SignerAdapter extends ethers.Signer {
     throw new Error("changing providers on a SignerAdapter is unsupported");
   }
 
-  async getAddress(): Promise<string> {
-    return ethers.utils.computeAddress(SecP256K1.UncompressedPoint.from(await this._isolatedKey.getPublicKey()));
+  async getAddress(addressNList: core.BIP32Path): Promise<string> {
+    const addressNode = await this.rootNode.derivePath(core.addressNListToBIP32(addressNList));
```
kaladinlight, line 39: `addressNode` -> `node` might be a bit better vernacular. No reason these nodes couldn't be derived at any level of the path and just more generic

packages/hdwallet-native/src/ethereum.ts
```diff
@@ -53,32 +54,35 @@ export function MixinNativeETHWallet<TBase extends core.Constructor<NativeHDWall
...
     }
 
     ethWipe() {
       this.#ethSigner = undefined;
     }
 
     async ethGetAddress(msg: core.ETHGetAddress): Promise<string | null> {
-      if (!_.isEqual(msg.addressNList, core.bip32ToAddressNList("m/44'/60'/0'/0/0"))) {
-        throw new Error("path not supported");
+      if (msg.addressNList.length < 5) {
```
kaladinlight, line 70: wondering if we should just omit this check and allow addresses to be derived in non standard ways if desired. we don't really enforce this anywhere else either

asamere: yep good point

packages/hdwallet-native/src/crypto/isolation/adapters/ethereum.ts
asamere: `rootNode` was definitely the wrong name. I changed it to master which I think reads better and better reflects what the variable is. lmk

kaladinlight: parent/child vernacular could be solid as well, parent is passed through (may be root or lower depth) and a child node is derived via the provided path

asamere: in the end @pastaghost wins the naming contest with `nodeAdapter`

0xdef1cafe (approved):

gomesalexandre (commented):
Conceptual stamp, logic looks sane, untested for now

gomesalexandre (approved):
Tested with linked hdwallet packages, confirmed derivation is working and producing the same address as when adding an Ethereum account with KK

PR #572, opened 2022-09-13
refactor: replace lodash isobject function

pastaghost:
The `isObject()` function from lodash causes runtime errors in the ShapeShift MetaMask snap. This PR replaces `isObject()` in hdwallet-native

### DIFF
--- a/packages/hdwallet-core/src/utils.ts
+++ b/packages/hdwallet-core/src/utils.ts
@@ -254,7 +254,7 @@ export function isIndexable(x: unknown): x is Record<string | number | symbol, u
   return x !== null && ["object", "function"].includes(typeof x);
 }
 
-/* Replacement for _.isObject() */
-export function isObject(x: any) {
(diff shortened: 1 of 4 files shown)


packages/hdwallet-core/src/utils.ts
```diff
@@ -253,3 +253,8 @@ export function compatibleBufferConcat(list: Uint8Array[]): Buffer {
 export function isIndexable(x: unknown): x is Record<string | number | symbol, unknown> {
   return x !== null && ["object", "function"].includes(typeof x);
 }
+
+/* Replacement for _.isObject() */
+export function isObject(x: any) {
+  return typeof x === "object" && x !== null;
+}
```
kaladinlight, line 260: 1:1 replacement assuming there wasn't an issue with the function typeof check?
```suggestion
/* Replacement for lodash.isObject() */
export function isObject(x: unknown) {
  return x !== null && ["object", "function"].includes(typeof x);
}
```

PR #573, opened 2022-09-13
feat: allow native wallet to be loaded with non-root node

pastaghost:
Feature:

HDWallet-native assumes that the node used to initialize a signer lives at the root of the key derivation tree. This PR adds an optional argument to `Node.create()`, allowing the user to explicitly specify the node's path in the tree.

Testing:

This feature does touch sensitive areas of the HDWallet-native code, but since the added parameter is optional, this is not expected to break existing applications consuming HDWallet-native.


kaladinlight (approved):
:fire:

0xdef1cafe (approved):

PR #574, opened 2022-09-16
feat: Support AVAX (and other chain signing) on keepkey

foufrix:
Support signing on other chains than Ethereum with Keepkey.

I have added a test with AVAX signing. My method for integrating the test and running it with KeepKey is not the best (all wallet will show success with AVAX but will not run the test). If you have a better way to implement with Jest?
Here is a resume of the problem I encounter with Jest: https://stackoverflow.com/questions/73746191/run-a-test-in-jest-only-if-certain-condition-are-met


foufrix:
found an elegant way here : https://github.com/facebook/jest/issues/7245#issuecomment-791285252

pastaghost:
@foufrix 

I pulled down the master branch of your fork, rebased your changes on top of the current shapeshift:hdwallet master branch, and added test data for the other wallet packages so that we don't have to run the test conditionally for KeepKey only. Can you give me permissions to push to the master branch of your hdwallet fork directly?

cc: @BitHighlander

foufrix:
@pastaghost of course you can go ahead!

PR #583, opened 2022-10-21
fix: use 60 as a slip44 for Avalanche

gomesalexandre:
SLIP44 should be 60 for all EVM chains, else wrong accounts will be derived.
9000 is the SLIP44 of Avalanche X-chain, which isn't an EVM chain.


0xdef1cafe (approved):

integration/src/ethereum/ethereum.ts
```diff
@@ -249,8 +249,12 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
         if (!wallet) return;
         // eslint-disable-next-line jest/no-jasmine-globals, jest/no-disabled-tests
 
+        if (ledger.isLedger(wallet)) return; // FIXME: just test kk and native for now
+        if (trezor.isTrezor(wallet)) return; // FIXME: just test kk and native for now
+        if (portis.isPortis(wallet)) return; // FIXME: just test kk and native for now
```
gomesalexandre, line 254: Nothing new here, if you open this module, we consistently exclude these including for Ethereum, I reckon the rationale is these aren't properly implemented yet/anymore?

kaladinlight: Huh, ok, well thanks for the info. Will approve as is then

kaladinlight (approved):

PR #586, opened 2022-10-27
feat: walletconnect to dapps

0xdef1cafe:
**needs package bump before merging**


packages/hdwallet-walletconnect-bridge/src/bridge.ts
```diff
@@ -0,0 +1,178 @@
...
+    this.options?.onCallRequest(payload);
+  }
+
+  public async approveRequest(
+    request: WalletConnectCallRequest,
+    approveData?: Partial<WalletConnectCallRequestResponseMap[keyof WalletConnectCallRequestResponseMap]>
+  ) {
+    let result: any;
+    switch (request.method) {
+      // TODO: figure out what to do for "eth_sign" and "eth_signTypedData".
```
0xdef1cafe, line 99: this is a problem and needs to be addressed

packages/hdwallet-walletconnect-bridge/src/bridge.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  describe("on session_request", () => {
+    it("should approve session with current chain and account", async () => {
+      await bridge._onSessionRequest(null, { params: [{ chainId: 123 }] });
+      expect(connector.approveSession).toHaveBeenCalledWith({
+        chainId: 123,
+        accounts: ["test address"],
+      });
+    });
+
+    it("should default to rinkeby chain", async () => {
```
0xdef1cafe, line 60: shapeshift does not, and never will, support rinkeby - we need to remove this default fallback case, and make sure all consumers explicitly specific a chain id when connecting a session

stackedq:
closing - we're moving these codes directly into the web repo.

PR #587, opened 2022-11-11
feat(hdwallet-keepkey): add eip-712 signtypeddata support

pastaghost:


elmutt (dismissed):

packages/hdwallet-keepkey/src/transport.ts
```diff
@@ -92,7 +92,13 @@ export class Transport extends core.Transport {
     const buffer = new Uint8Array(9 + 2 + msgLength);
     buffer.set(first.slice(0, Math.min(first.length, buffer.length)));
 
-    for (let offset = first.length; offset < buffer.length; ) {
+    // msgLength includes the preamble overhead "?" for each chunk.
+    // We don't want the preamble from each chunk in the buffer (though, curiously, the preamble 
+    // for the first chunk is included in the buffer for some reason).
+    // We do, however, have to count the preamble in each chunk to make the proper number
+    // of chunked reads
+    let msgLenReceived = 64 // already have first buffer
```
BitHighlander, line 100: 👏

BitHighlander (dismissed):
exciting 🚢

packages/hdwallet-keepkey/src/bitcoin.ts
```diff
@@ -70,36 +70,7 @@ function prepareSignTx(
     const output: core.BTCSignTxOutput = o;
     const newOutput = new Types.TxOutputType();
     newOutput.setAmount(Number(output.amount));
-    if (output.exchangeType) {
```
kaladinlight, line 73: seems like unrelated changes. was this related to legacy shapeshift exchange?

packages/hdwallet-keepkey/src/bitcoin.ts
```diff
@@ -112,6 +83,7 @@ function prepareSignTx(
     } else {
       // BTCSignTxOutputSpend
       newOutput.setScriptType(Types.OutputScriptType.PAYTOADDRESS);
+      assert(output.address != undefined, "Output must have a valid BTC address.");
```
kaladinlight, line 86: preference for strict equality checks. could also just do a falsey `!output.address` check as it is really checking if value exists and not necessarily if valid or not (could update error string to reflect this as well).
```suggestion
      assert(output.address !== undefined, "Output must have a valid BTC address.");
```

packages/hdwallet-keepkey/src/bitcoin.ts
kaladinlight: if not supported anymore, I would clean up any related types as well and just make a separate PR for this removal

packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -1,14 +1,16 @@
+/* eslint-disable no-debugger */
```
kaladinlight, line 1:
```suggestion
```

integration/src/ethereum/ethereum.ts
```diff
@@ -421,5 +423,54 @@ export function ethereumTests(get: () => { wallet: core.HDWallet; info: core.HDW
       },
       TIMEOUT
     );
+
+    test.skip(
```
kaladinlight, line 427:
```suggestion
    test(
```

integration/src/ethereum/ethereum.ts
```diff
@@ -1,4 +1,6 @@
+/* eslint-disable jest/no-disabled-tests */
```
kaladinlight, line 1:
```suggestion
```

packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -188,21 +173,116 @@ export async function ethGetAddress(transport: Transport, msg: core.ETHGetAddres
...
+  transport: Transport,
+  msg: core.ETHSignTypedData
+): Promise<core.ETHSignedTypedData> {
+  /**
+   * If the message to be signed is sufficiently small, the KeepKey can calculate the
+   * domain separator and message hashes. Otherwise, we need to pre-calculate hashes
+   * here and verify on device.
+   */
+
+  let result: Ethereum.EthereumTypedDataSignature;
```
kaladinlight, line 199: not being reassigned within logic forks, may as well just assign as const locally scoped

packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -188,21 +173,116 @@ export async function ethGetAddress(transport: Transport, msg: core.ETHGetAddres
...
+
+  const sTypes = JSON.stringify({ types: msg.types });
+  const sPrimaryType = JSON.stringify({ primaryType: msg.primaryType });
+  const sDomain = JSON.stringify({ domain: msg.domain });
+  const sMessage = JSON.stringify({ message: msg.message });
+
+  try {
+    if (sTypes.length > 2048 || sPrimaryType.length > 80 || sDomain.length > 2048 || sMessage.length > 2048) {
+      /* Pre-calculate domain separator and messages hashes and verify on KeepKey */
+      const hashableMessage: TypedData = {
```
kaladinlight, line 209: thoughts about nesting this object in `coreETHSignTypedData` to prevent the need to recreate the `TypedData` object?
```
export type ETHSignTypedData = { data: TypedData; addressNList: BIP32Path };
```

packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -188,21 +173,116 @@ export async function ethGetAddress(transport: Transport, msg: core.ETHGetAddres
...
+      let messageHash = undefined;
+      const domainSeparatorHash = getTypeHash(hashableMessage, "EIP712Domain");
+      const domainSeparatorHash64 = Buffer.from(domainSeparatorHash).toString("base64");
+
+      const t = new Ethereum.EthereumSignTypedHash();
+      t.setAddressNList(msg.addressNList);
+      t.setDomainSeparatorHash(domainSeparatorHash64);
+
+      if (msg.message) {
+        messageHash = getMessage(hashableMessage, true);
```
kaladinlight, line 224: message is required on `TypedData` and while testing, it looks like `getTypeHash` throws if the payload does not conform correctly as is, don't think we need this check and can get rid of the `let` above

packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -188,21 +173,116 @@ export async function ethGetAddress(transport: Transport, msg: core.ETHGetAddres
...
+      let response = await transport.call(Messages.MessageType.MESSAGETYPE_ETHEREUM712TYPESVALUES, dsh, {
+        msgTimeout: core.LONG_TIMEOUT,
+        omitLock: true,
+      });
+
+      const mh = new Ethereum.Ethereum712TypesValues();
+      mh.setAddressNList(msg.addressNList);
+      mh.setEip712types(sTypes);
+      mh.setEip712primetype(sPrimaryType);
+
```
kaladinlight, line 260:
```suggestion
```

packages/hdwallet-keepkey/src/ethereum.ts
```diff
@@ -188,21 +173,116 @@ export async function ethGetAddress(transport: Transport, msg: core.ETHGetAddres
...
+      mh.setEip712data(sMessage);
+      mh.setEip712typevals(2);
```
kaladinlight, line 262:
```suggestion
      mh.setEip712typevals(2);

```

packages/hdwallet-keepkey/src/transport.ts
```diff
@@ -92,7 +92,13 @@ export class Transport extends core.Transport {
...
     buffer.set(first.slice(0, Math.min(first.length, buffer.length)));
 
-    for (let offset = first.length; offset < buffer.length; ) {
+    // msgLength includes the preamble overhead "?" for each chunk.
+    // We don't want the preamble from each chunk in the buffer (though, curiously, the preamble
+    // for the first chunk is included in the buffer for some reason).
+    // We do, however, have to count the preamble in each chunk to make the proper number
+    // of chunked reads
+    let msgLenReceived = 64; // already have first buffer
+    for (let offset = first.length; msgLenReceived < msgLength; msgLenReceived += 64) {
```
kaladinlight, line 101: weird ordering of the for loop vars
```suggestion
    let offset = first.length;
    for (let msgLenReceived = 64; msgLenReceived < msgLength; msgLenReceived += 64) {
```

packages/hdwallet-keepkey/src/ethereum.ts
pastaghost: sTypes, sPrimaryType, sDomain, sMessage are representations of the data in the ETHSignedTypedData object formatted specifically for KeepKey. Personally, I like to keep application-specific logic as close to the application as possible. Here, I think it's cleaner to transform the data for KeepKey in the hdwallet-keepkey adapter instead of extending the shared data type to include fields that are only consumed in one place.

kaladinlight: Just to clarify, I was meaning this comment to be in reference to `hashableMessage` vs the `sVariables`. 
image

pastaghost: We aren't yet using this type anywhere else, and there isn't any reason why the object necessarily needs to be flat, so I don't see any issue with `core.ETHSignTypedData` having separate fields for the hashable data and addressNList. I'll make that change now.

kaladinlight: Feels more ergonomic this way to me. Cool, thanks!

examples/sandbox/index.ts
```diff
@@ -1819,6 +1822,36 @@ $ethEIP1559.on("click", async () => {
...
 
+$ethSignTypedData.on("click", async (e) => {
+  e.preventDefault();
+  if (!wallet) {
+    $ethResults.val("No wallet?");
+    return;
+  }
+  if (core.supportsETH(wallet)) {
+    const result = await wallet.ethSignTypedData({
+      ...eip712["calculateHashesOnDevice"],
```
kaladinlight, line 1833: will have to update the sandbox payloads for `hashableData` as well so we don't forget

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -107,6 +108,18 @@ export interface ETHVerifyMessage {
   signature: string;
 }
 
+export type ETHSignTypedData = {
+  hashableData: TypedData;
+  addressNList: BIP32Path;
+};
+
+export type ETHSignedTypedData = {
```
kaladinlight, line 116: @pastaghost is there a reason for returning the `domainSeparatorHash` and `messageHash` as opposed to just the address + sig like in `signMessage`?

PR #596, opened 2023-02-06
feat: add bsc

kaladinlight:
- `supportsBSC` flag


gomesalexandre (dismissed):

gomesalexandre (approved):

packages/hdwallet-core/src/utils.ts
```diff
@@ -152,6 +152,7 @@ const slip44Table = Object.freeze({
   Ethereum: 60,
   Avalanche: 60,
   Optimism: 60,
+  BnbSmartChain: 60,
```
pastaghost, line 155: Small nitpick - maybe `BinanceSmartChain` instead? I see both used online - so this is just an aesthetic thing.

pastaghost (approved):

packages/hdwallet-core/src/utils.ts
kaladinlight: for visibility from dms:
https://www.binance.com/en/blog/ecosystem/introducing-bnb-chain-the-evolution-of-binance-smart-chain-421499824684903436

image

PR #598, opened 2023-02-25
fix: xdefi _supportsOptimism false

gomesalexandre:
#### Description

XDefi doesn't support Optimism, see https://www.xdefi.io/article/multichain-wallet-xdefi-support/


kaladinlight:
image

their docs are out of date it looks like. I am seeing optimism in the extension.

PR #601, opened 2023-03-03
feat: update tallyho capabilities

gomesalexandre:
See the supported chains in latest TallyHo stable

image

Confirmed both contract calls on target chain and `ethSwitchChain` are working (interestingly, in a silent way similar to how native wallet handles multiple chains), see testing while connected to Ethereum:

image
image
image


packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -256,7 +256,7 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
       // https://docs.metamask.io/guide/ethereum-provider.html#errors
       // Internal error, which in the case of wallet_switchEthereumChain call means the chain isn't currently added to the wallet
       if (error.code === -32603) {
-        // TODO: TallyHo currently supports a finite number of chains natively (Mainnet + Polygon/Arbitrum/Optimism under feature flag), with no capabilities to add new chains
+        // TODO: TallyHo currently supports a finite number of chains natively, with no capabilities to add new chains
```
gomesalexandre, line 259: Support should be here soonish https://github.com/tahowallet/extension/blob/d5811da18605927dda92280092f4d41f3e432d58/background/services/internal-ethereum-provider/index.ts#L267

packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -88,9 +88,9 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
   readonly _supportsETH = true;
   readonly _supportsETHInfo = true;
   readonly _supportsEthSwitchChain = false;
```
gomesalexandre, line 90: Apparently supported, though I didn't get this to work using Tally with other dApps
https://github.com/tahowallet/extension/blob/d5811da18605927dda92280092f4d41f3e432d58/background/services/internal-ethereum-provider/index.ts#L289-L297

Will give this some more testing and PR the capability update in a follow-up PR after confirming this actually does what it says on the box

PR #604, opened 2023-03-13
handle pin mismatch

BitHighlander:
This patch sending the expected pin code for the handled case of a user attempting to set a new pin with mismatching pins. This patch is required for proper keepkey onboarding.


millsmcilroy (approved):
lgtm

kaladinlight:
@BitHighlander please run `yarn lerna:version-patch` and we can get this in. Thanks!

PR #606, opened 2023-03-13
add transport lock to ethSignTypedData

BitHighlander:
Transport locking is needed for any request where a user must interact with the device. This pr adds transport locking to ethSignTypedData and prevents other requests from breaking the sign eip712 tx flow


kaladinlight:
@BitHighlander please rebase against master and run `yarn lerna:version-patch` accordingly and then we can get this in. Thank you!

PR #610, opened 2023-04-11
Add AVAX EIP-1559 reference TX to match SS prod

BitHighlander:
This pr adds a integration test to reference how shapeshift prod manages AVAX sending currently.


0xean:
@kaladinlight to handle this internally.

PR #612, opened 2023-04-14
fix: upgrade lerna

millsmcilroy:
`lerna` depends on `node-gyp` and the old version of `node-gyp` does not work with Python 3 ([File mode U is deprecated in Python 3](https://github.com/nodejs/gyp-next/pull/83)), causing `yarn install` to fail:
```
ValueError: invalid mode: 'rU' while trying to load binding.gyp
gyp ERR! configure error
gyp ERR! stack Error: `gyp` failed with exit code: 1
```


millsmcilroy:
Closing due to lack of reviews/feedback

PR #617, opened 2023-05-02
feat(hdwallet-native): add arkeo support

pastaghost:


packages/hdwallet-native/src/arkeo.ts
```diff
@@ -0,0 +1,99 @@
...
+import * as core from "@shapeshiftoss/hdwallet-core";
+import * as bech32 from "bech32";
+import CryptoJS from "crypto-js";
+import PLazy from "p-lazy";
+
+import * as Isolation from "./crypto/isolation";
+import { NativeHDWalletBase } from "./native";
+import * as util from "./util";
+
+const ARKEO_CHAIN = "arkeo-mainnet";
```
kaladinlight, line 12: do we know for sure what the mainnet chain id is going to be? this makes sense enough, but just checking

packages/hdwallet-native/src/arkeo.test.ts
```diff
@@ -0,0 +1,88 @@
...
+              amount: "2291",
+              denom: "uarkeo",
+            },
+          ],
+          gas: "91633",
+        },
+        signatures: [],
+        memo: "foobar",
+      },
+      chain_id: "arkeo-1",
```
kaladinlight, line 79: seeing `arkeo-1` chain ids that will need to match up with desired mainnet chain id

examples/sandbox/json/arkeo/arkeoAminoTx.json
```diff
@@ -0,0 +1,258 @@
...
+          "creator": "arkeo15cenya0tr7nm3tz2wn3h3zwkht2rxrq7r7z5sh",
+          "to_address": "arkeo15cenya0tr7nm3tz2wn3h3zwkht2rxrq7r7z5sh"
+        }
+      }
+    ],
+    "sequence": "90"
+  },
+  "arkeoAddClaimTx": {
+    "account_number": "97721",
+    "chain_id": "arkeo-1",
```
kaladinlight, line 234: same chain id comment in these reference txs

examples/sandbox/index.ts
```diff
@@ -1136,6 +1147,17 @@ const $thorchainLiquidityAmount = $("#thorchainLiquidityAmount");
...
+const $thorchainArkeoBondProvider = $("#thorchainArkeoBondProvider");
+const $thorchainArkeoModProvider = $("#thorchainArkeoModProvider");
+const $thorchainArkeoOpenContract = $("#thorchainArkeoOpenContract");
+const $thorchainArkeoCloseContract = $("#thorchainArkeoCloseContract");
+const $thorchainArkeoClaimContractIncome = $("#thorchainArkeoClaimContractIncome");
+const $thorchainArkeoClaimEth = $("#thorchainArkeoClaimEth");
+const $thorchainArkeoClaimArkeo = $("#thorchainArkeoClaimArkeo");
+const $thorchainArkeoTransferClaim = $("#thorchainArkeoTransferClaim");
+const $thorchainArkeoAddClaim = $("#thorchainArkeoAddClaim");
+const $thorchainArkeoResults = $("#thorchainArkeoResults");
```
kaladinlight, line 1159: remove thorchain vernacular

examples/sandbox/index.ts
```diff
@@ -1506,6 +1528,195 @@ $thorchainSignAddLiquidity.on("click", async (e) => {
   }
 });
 
+$thorchainArkeoBondProvider.on("click", async (e) => {
+  e.preventDefault();
+  if (!wallet) {
+    $thorchainNativeResults.val("No wallet?");
+    return;
+  }
+  if (core.supportsThorchain(wallet)) {
```
kaladinlight, line 1537: use narrow as arkeo wallet throughout

examples/sandbox/index.ts
```diff
@@ -1506,6 +1528,195 @@ $thorchainSignAddLiquidity.on("click", async (e) => {
...
+    const res = await wallet.thorchainSignTx({
+      addressNList: core.bip32ToAddressNList(`m/44'/931'/0'/0/0`),
+      chain_id: "thorchain",
```
kaladinlight, line 1540: update all paths/chain_id references to arkeo

examples/sandbox/index.ts
pastaghost: This is relevant because it's in the THORChain block, separate from the Arkeo block. We need to be able to sign Arkeo transactions on THORChain, so this is in place to validate that. 

Effectively, this comes down to switching the chain ID, so I could always remove this from the THORChain block and add a chain ID selector to the Arkeo block and accomplish the same thing.

examples/sandbox/index.ts
pastaghost: See above

examples/sandbox/index.ts
kaladinlight: Oh I did not make it down far enough to see the arkeo block and though it was copy/pasta issue. And signing of arkeo txs on thorchain is an aspect I was not aware of tbh.

I don't know what would make more sense from an organization standpoint since it is kind of a different and strange use case. As it is the sandbox, I would just leave it as is.

0xdef1cafe (approved):

PR #618, opened 2023-05-11
chore: gnosis support

lmyslinski:
Adds Gnosis chain support to HDWallet


packages/hdwallet-tallyho/src/tallyho.ts
```diff
@@ -91,6 +91,7 @@ export class TallyHoHDWallet implements core.HDWallet, core.ETHWallet {
   readonly _supportsAvalanche = false;
   readonly _supportsOptimism = false;
   readonly _supportsPolygon = true;
+  readonly _supportsGnosis = true;
```
kaladinlight, line 94: not supported out of the box
```suggestion
  readonly _supportsGnosis = false;
```

packages/hdwallet-xdefi/src/xdefi.ts
```diff
@@ -93,6 +93,7 @@ export class XDEFIHDWallet implements core.HDWallet, core.ETHWallet {
   readonly _supportsOptimism = true;
   readonly _supportsBSC = true;
   readonly _supportsPolygon = true;
+  readonly _supportsGnosis = true;
```
kaladinlight, line 96: also not listed as supported
```suggestion
  readonly _supportsGnosis = false;
```
image

packages/hdwallet-tallyho/src/tallyho.ts
gomesalexandre: Actually, fresh from a few hours ago, should be safe to add

> You can now use any EVM network listed on https://chainlist.org/ inside Taho! Including zksync, fantom, canto, celo, gnosis, moonbeam, etc. 🎉 

image

But yeah either or, we don't really support this wallet

gomesalexandre (approved):

packages/hdwallet-tallyho/src/tallyho.ts
gomesalexandre: Also this would require turning on ethSwithChain support for Tally, which doesn't matter since we've removed it as an option in the app 👋🏽

PR #635, opened 2023-09-28
fix: `ethSignMessage` expect hex string message

0xApotheosis:
See https://github.com/shapeshift/web/pull/5382 for the web implementation - this updates types / implementations across HDWallet packages to always expect an hex string as a message input - *not* a bytes or string union, which we were wrongly handling and doing one too many/one too little conversion when doing the string to Uint8Bytes to Buffer to hex dance.

Tested in https://github.com/shapeshift/web/pull/5382 with Verdaccio-published packages:

- WC V2

image

- Native

image

- MM Multichain snap

image

- MM

image

- Ledger

Currently WIP and borked e2e, but confirmed the signature bits are working:

image

- KK

image

-----

The currently implementation seems wrong.

When we sign a message, and then recover the message address from the signed message, the address does not match the signer.

This PR fixes that.


packages/hdwallet-walletconnectV2/src/ethereum.ts
```diff
@@ -87,11 +86,9 @@ export async function ethSignMessage(
   args: { data: string | Bytes; fromAddress: string },
```
gomesalexandre, line 86: If we remove the string to buffer to hex string dance, we'll have to be 100% sure this comes as a string
```suggestion
  args: { data: string; fromAddress: string },
```
See https://docs.metamask.io/wallet/reference/personal_sign

packages/hdwallet-walletconnectV2/src/ethereum.ts
```diff
@@ -87,11 +86,9 @@ export async function ethSignMessage(
...
   provider: EthereumProvider
 ): Promise<ETHSignedMessage | null> {
-  const buffer = isBytes(args.data) ? Buffer.from(arrayify(args.data)) : Buffer.from(args.data);
-
```
gomesalexandre, line 91: Assuming this now comes as an hex string, we may want to use https://docs.ethers.org/v5/api/utils/bytes/#utils-isHexString to guard at runtime against non-hexlified strings being passed

gomesalexandre:
Note - `yarn lerna:version-patch` will need to be ran here - since we have multiple hdwallet PRs in flight, this could run things here, make sure to run this as a last step right before merging

examples/sandbox/package.json
gomesalexandre, line 1: Revert this commit before merging me

0xdef1cafe (approved):

PR #642, opened 2023-10-08
fix: support chain switching for walletconnect

woodenfurniture:
Updates WalletConnect v2 implementation to support chain switching.
See https://github.com/shapeshift/web/pull/5425 for details.


packages/hdwallet-walletconnectV2/src/walletconnectV2.ts
```diff
@@ -358,4 +358,22 @@ export class WalletConnectV2HDWallet implements HDWallet, ETHWallet {
...
+  }
+
+  public async ethSwitchChain({ chainId }: AddEthereumChainParameter): Promise<void> {
+    const parsedChainId = parseInt(chainId, 16);
+    if (isNaN(parsedChainId) || this.chainId === parsedChainId) {
+      return;
+    }
+
+    this.chainId = parsedChainId;
+    this.provider.chainId = parsedChainId;
```
gomesalexandre, line 373: Do we need to mutate this optimistically? `this.provider` being the Ethereum provider, its `chainId` property should automagically reflect the one after the switch?

packages/hdwallet-walletconnectV2/src/walletconnectV2.ts
```diff
@@ -358,4 +358,22 @@ export class WalletConnectV2HDWallet implements HDWallet, ETHWallet {
...
+    return this.provider.chainId;
+  }
+
+  public async ethSwitchChain({ chainId }: AddEthereumChainParameter): Promise<void> {
+    const parsedChainId = parseInt(chainId, 16);
+    if (isNaN(parsedChainId) || this.chainId === parsedChainId) {
+      return;
+    }
+
+    this.chainId = parsedChainId;
```
gomesalexandre, line 372: We may want to not be optimistic here and move this after the promise execution

packages/hdwallet-walletconnectV2/src/walletconnectV2.ts
woodenfurniture: excellent point, actioning now

gomesalexandre (dismissed):
Tested with https://github.com/shapeshift/web/pull/5425 - stamp pending lerna version bumpy

gomesalexandre (approved):

PR #654, opened 2023-11-21
perf: remove argon bench on boot, use sane defaults across the board

woodenfurniture:
## Overview

Hardcodes sensible defaults for argon encryption parameters, skipping the benchmark to reduce initial app boot time by >4s on both desktop and mobile devices, when connecting to native wallet for the first time.

The default parameters remain the same, with only the `iterations` changing from 23 to 16. Noting that reasonably modern mobile devices achieve ~11 iterations within the 1s target, the default was set to a conservative 16 iterations as a balance between performance and security. 

## Testing

This should be tested locally with shapeshift web via local npm registry (e.g [verdaccio](https://verdaccio.org/))
- **can open native wallet with existing local state** (existing parameters set in IndexedDB with prior benchmark on prior version version). This should prove backwards compatibility so users don't have to re-import their wallets
- **can open native wallet starting from empty local state** (clear IndexedDB)

## Issue

Will close [#5557](https://github.com/shapeshift/web/issues/5557) after bumping version in web.

### Benchmark parameters

**Mobile - Pixel 4a**
Note overall duration of the bench was 4.1s
image

**Desktop - Mac M2**
Note overall duration of the bench was 4.6s
image

### Performance results (on Desktop - Mac M2)

**Before changes**
4s benchmark runs on first open of native wallet:
image

**After changes**
4s benchmark no longer run:
image

Note the defaults get stored as intended:
image


examples/sandbox/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shapeshiftoss/hdwallet-sandbox",
-  "version": "1.52.12",
+  "version": "1.52.14",
```
gomesalexandre, line 3: Any reason why we skipped 1.52.13 and went straight into 1.52.14?

woodenfurniture: the script failed locally when incrementing, saying 13 already exists

gomesalexandre: Will fix that for you 🙏🏽

gomesalexandre:
Current dependencies on/for this PR:
* master
  * **PR #654** <a href="https://app.graphite.dev/github/pr/shapeshift/hdwallet/654?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/shapeshift/hdwallet/654?utm_source=stack-comment).

examples/sandbox/package.json
gomesalexandre: ac6b7a44aaeaefa47d1521131a54da4b4a77c874

0xApotheosis (approved):
Solid big wood

PR #656, opened 2023-12-08
fix: mm request permissions on request accounts pending

gomesalexandre:
See https://github.com/MetaMask/metamask-extension/issues/10085 for reference

tl;dr is MM API doesn't make any sense, and doesn't allow for `eth_requestAccounts` requests if there's already a pending one. This will throw errors and make users unable to unlock their wallet, effectively being in an infinite loop.

Having request/s pending could happen if users dismissed previous connection requests and/or wallet unlock, or worse, if MM popup never properly popped up and discretely went with a notification instead (which can sometimes happen and is out of our control).

This PR uses `wallet_requestPermissions` instead when the former failed. That will ensure MM properly pops up when it should, the downside of this being users will have to "reconnect" accounts even if they're already connected.

### Screenshots

(Don't mind the slowness / snap UI missing, this was with localhost debugging)

https://github.com/shapeshift/hdwallet/assets/17035424/6040e2e5-1017-4ff7-8f3d-7d63c4a044c7

### Issue

Relates to web's https://github.com/shapeshift/web/issues/5453


0xApotheosis (approved):
This makes a really annoying UX moderately annoying 🙏 

Played with `wallet_requestPermissions` and confirmed it does what we want it to do by forcing the Metamask window to open (albeit forcing the user to reselect accounts, but it's a definite improvement nonetheless!).

PR #670, opened 2024-05-09
fix: support cosmos and thorchain when validating current app on ledger device

woodenfurniture:
Adds support for checking the status of the current app on ledger for Cosmos and THORChain.

Removes some dead code.

Relates to https://github.com/shapeshift/web/issues/6806


packages/hdwallet-ledger/src/utils.ts
```diff
@@ -96,503 +95,277 @@ export const createXpub = (
...
+  2: "Litecoin",
+  3: "Dogecoin",
+  5: "Dash",
+  20: "Digibyte",
+  60: "Ethereum",
+  118: "Cosmos",
+  145: "Bitcoin Cash",
+  931: "THORChain",
+};
+
```
woodenfurniture, line 365: THORChain and Cosmos do not require any metadata from `networksUtil`, so it made no sense to add an entry there for the sake of enabling app name lookup. Instead, I've broken it out into the lookup above.

packages/hdwallet-ledger/src/utils.ts
```diff
@@ -96,503 +95,277 @@ export const createXpub = (
...
     };
-    messagePrefix: string;
-    pubKeyHash?: number;
-    scriptHash?: number;
-    wif?: number;
   };
   sigHash?: number;
   isSegwitSupported?: boolean;
-  handleFeePerByte?: boolean;
-  additionals?: string[];
```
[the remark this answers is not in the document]
gomesalexandre: We may not use them now and probably never in the future, but would keep it just to avoid silently breaking things - according to grepping ledger repos, this is perfectly valid (though possibly outdated and not relevant anymore, but who knows...) https://github.com/LedgerHQ/ledger-wallet-webtool/blob/2db847a27e42b2353e979240cc711ad8fb46fe71/src/Networks.js#L159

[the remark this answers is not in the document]
gomesalexandre: Regardless of whether or not it is used/anymore, it does conforms to *some* schema (no matter how outdated it may be), which confirms we probably shouldn't clean this up to keep on honoring said schema

gomesalexandre (approved):
Tested locally, confirmed this does what it says on the box

https://github.com/shapeshift/hdwallet/assets/17035424/5d9b13db-e89e-4d49-bf00-ba2d0750cfcb

packages/hdwallet-ledger/src/utils.ts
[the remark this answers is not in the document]
gomesalexandre: On second thought: meh

gomesalexandre (dismissed):

examples/sandbox/package.json
```diff
@@ -11,21 +11,21 @@
...
-    "@shapeshiftoss/hdwallet-ledger-webusb": "1.53.6",
-    "@shapeshiftoss/hdwallet-metamask": "1.53.6",
-    "@shapeshiftoss/hdwallet-native": "1.53.6",
-    "@shapeshiftoss/hdwallet-portis": "1.53.6",
-    "@shapeshiftoss/hdwallet-tallyho": "1.53.6",
-    "@shapeshiftoss/hdwallet-trezor": "1.53.6",
-    "@shapeshiftoss/hdwallet-trezor-connect": "1.53.6",
-    "@shapeshiftoss/hdwallet-walletconnect": "1.53.6",
-    "@shapeshiftoss/hdwallet-xdefi": "1.53.6",
+    "@shapeshiftoss/hdwallet-core": "^1.53.7-alpha.0",
```
kaladinlight, line 14: still pointing to alpha packages

packages/hdwallet-ledger/src/utils.ts
```diff
@@ -96,503 +95,277 @@ export const createXpub = (
...
   bitcoinjs: {
-    bech32?: string;
     bip32: {
-      private?: number;
       public: Partial<Record<core.BTCInputScriptType, number>>;
     };
-    messagePrefix: string;
-    pubKeyHash?: number;
-    scriptHash?: number;
-    wif?: number;
```
kaladinlight, line 113: dangerous removal, please revert bitcoinjs changes. we are passing this payload to bitcoinjs for transaction building which is using at least `bech32`, `pubKeyHash` and `scriptHash` at first glance and likely others.

Happy to address this in a follow up PR to update ledger to use modern psbt and pull out bitcoinjs network configs into a base package for use against any packages using bitcoinjs-lib.

would be fine with removal of the unused values outside of bitcoinjs being removed at this time

gomesalexandre: Cleaned diff up to the bare minimum this is bringing in https://github.com/shapeshift/hdwallet/pull/670/files#diff-48c2229f8d069d68109d331ff81dba7fc4d501ae143cb400cae2b543dda1d119 - @woodenfurniture FYI ^^

kaladinlight (approved):

gomesalexandre (approved):

PR #676, opened 2024-08-30
feat: wcv2 eip712 signing

gomesalexandre:
### Screenshots

image

### Description

Does what it says on the box - note that this does *not* work for sc wallets, only EOAs.

For EOAs, we will need to tackle that case specifically, since eip-712 signing doesn't work on the CoW side, and things are a bit different there (requires both an on-chain Tx *and* a signature):

https://docs.cow.fi/cow-protocol/reference/core/signing-schemes#presign

### Issue

- tackles https://github.com/shapeshift/web/issues/7587, web to follow


kaladinlight (approved):

gomesalexandre:
1 @kaladinlight hdwallet stamp = get in

PR #674, opened 2024-07-25
feat: real EIP-712 structured signing for Ledger

gomesalexandre:
### Description

This PR is a stretch of https://github.com/shapeshift/hdwallet/pull/673, authored as a separate PR, since it is inherently more risky:

- bumps all relevant Ledger deps to latest
- handles *actual* structured signing for Ledger vs. working, but ugly signing of hashed EIP-712 message

### Risks

Theoretically low, but given how Ledger types seem to break across releases, it's better to assume runtime is the same and full of breaking changes. Do a full regression testing of Ledger:

- Deriving accounts 0 across all chains types (EVM chains/UTXO chains/Cosmos/THOR)
- Deriving accounts 1+ across all chains types (EVM chains/UTXO chains/Cosmos/THOR)
- Signing *Txs* across all chains types (EVM chains/UTXO chains/Cosmos/THOR)
- Signing *messages* is happy for EVM chains, i.e wherever sign-in and CoW swaps 

### Testing

- On Nano S+ and X, CoW structured signing is now *real* structured signing, with typed data display (`chainId`, `version`, `name` etc) on-device vs. a dumb hash previously)
- On Nano S, CoW structured signing is actually not structured and the same as previusly (untested as I don't have a nano S handy for testing)
- On MM, EIP-712 structured signing is still happy

PR #681, opened 2024-09-11
feat: make hdwallet less pain

gomesalexandre:
### Description

a.k.a make sandbox compile fine with Ledger packages so we don't have to pin older ones anymore and can leverage newer Ledger package features.

Confirmed this builds/runs dev server fine in isolation as well as in https://github.com/shapeshift/hdwallet/pull/674.

### Testing

- Ensure `yarn && yarn build && examples/sandbox && yarn build` is happy
- Ensure sandbox `build`, `preview` and `clean` scripts are happy and the dev server outputs a valid build

### Screenshots

- Sandbox

image

- Happy web

image


.nvmrc
```diff
@@ -1 +1 @@
-16
+18
```
gomesalexandre, line 1: Required for vite and ensures we are aligned with web versions

examples/sandbox/index.html
```diff
@@ -320,7 +320,7 @@ <h4>ERC-20</h4>
       <button class="button button-outline" name="erc20Button" id="erc20TransferFrom">Transfer From</button>
       <center><p id="erc20Tooltip">Select an option from the above row. Click "submit" below to process.</p></center>
       <div class="erc20DynamicContainer" id="erc20DynamicContainer"></div>
-      <textarea disabled placeholder="Result" id="erc20Results" />
+      <textarea disabled placeholder="Result" id="erc20Results"></textarea>
```
gomesalexandre, line 323: html was rugged, dis not react but somehow was happy with parcel

packages/hdwallet-ledger/src/ethereum.ts
```diff
@@ -8,7 +8,7 @@ import EthereumTx from "ethereumjs-tx";
 // TODO: fix ts-ignore
 import * as ethereumUtil from "ethereumjs-util";
 import { arrayify, isBytes, joinSignature } from "ethers/lib/utils.js";
-import { isHexString } from "ethjs-util";
+import { isHexString } from "ethers/lib/utils";
```
gomesalexandre, line 11: May have worked at some point because of transient deps but definitely isn't now
 
image

examples/sandbox/package.json
```diff
@@ -5,13 +5,14 @@
...
-    "dev": "yarn parcel index.html",
-    "clean": "rm -rf dist node_modules public .parcel-cache"
+    "build": "rm -rf public && vite build --outDir public",
+    "preview": "vite",
+    "clean": "rm -rf dist node_modules public"
   },
   "dependencies": {
     "@metamask/eth-sig-util": "^7.0.0",
     "@shapeshiftoss/hdwallet-core": "1.55.2",
+    "@shapeshiftoss/hdwallet-keepkey": "1.55.2",
```
gomesalexandre, line 15: sandbox is derped and it working was accidental due to how parcel resolved modules, it was actually missing a ton and would stderr on build

packages/hdwallet-core/src/utils.ts
gomesalexandre, line 1: lint

packages/hdwallet-ledger-webhid/src/adapter.ts
gomesalexandre, line 1: Surfaced by the node update but was always there, no ledger diff in this PR just yet

packages/hdwallet-ledger/src/ethereum.ts
gomesalexandre, line 1: lint and import from correct path

packages/hdwallet-native/src/crypto/isolation/core/digest/types.ts
gomesalexandre, line 1: lint

.circleci/config.yml
```diff
@@ -5,10 +5,10 @@ orbs:
     executors:
       node:
         docker:
-          - image: cimg/node:16.20.2
+          - image: cimg/node:18.18.2
```
kaladinlight, line 8: feels like we can just use `18.20` LTS release?

packages/hdwallet-ledger-webhid/src/adapter.ts
kaladinlight: can follow the same pattern as `hdwallet-ledger-webusb/transport.ts` in `hdwallet-ledger-webhid/transport.ts` instead of the ts comment maybe
```
  constructor(device: HIDDevice, transport: TransportWebHID, keyring: core.Keyring) {
    super(transport as Transport, keyring);
```

.circleci/config.yml
gomesalexandre: The reasoning here was to keep the same as what `nvm use` yielded for me, though I probably had a cached 18.x version:

image

However I `nvm uninstall` and `nvm install`'d again and looks like 18.20.4 is the one that's going to be installed currently:

image

Might as well go with [current LTS](https://nodejs.org/download/release/latest-v18.x/)'s latest minor then! bb39b7972ba456de8c4613582ce8339b8b3d30bb

packages/hdwallet-ledger-webhid/src/adapter.ts
gomesalexandre: Can absolutely do: f6a1bfe7

Wondering if we just ditch webhid support in a follow-up? We can always bring it up if we deem fit, but seem to cause more hassle than it should all the while we don't support webhid 🤔

kaladinlight: Yeah have thought about the same thing and even portis/tallyho. Being open source I don't know if anyone else may be using them or if they even work as it stands tbh...

gomesalexandre:
Moving back to draft as CI passes and sandbox build *does* build but there are a few runtime issues there, will need to also tackle those dumb `require()` across hdwallet and convert to import - if some are still needed, we'll need to use the relevant [plugin](https://vitejs.dev/plugins/) e.g vite-plugin-commonjs / vite-plugin-require, but ideally we should avoid mixing ESM and CJS

examples/sandbox/vite.config.ts
```diff
@@ -0,0 +1,29 @@
+import { UserConfig } from "vite";
+import { nodePolyfills } from "vite-plugin-node-polyfills";
+
+const config: UserConfig = {
```
kaladinlight, line 4: :partying_face:

packages/hdwallet-ledger-webusb/src/transport.ts
```diff
@@ -81,22 +82,22 @@ export async function translateCoinAndMethod<T extends LedgerTransportCoinType,
 ): Promise<LedgerTransportMethod<T, U>> {
   switch (coin) {
     case "Btc": {
-      const btc = new Btc({ transport });
+      const btc = new Btc({ transport: transport as unknown as Transport });
```
kaladinlight, line 85: shouldn't need `unknown` cast first?

tsconfig.json
```diff
@@ -3,8 +3,8 @@
     // tsc is insanely drunk with internal resolutions currently, this at least silences it so we can ship things
     // the installed versions are still the same as before, but the bumped tsc version might have made it stricter
     "skipLibCheck": true,
-    "target": "ES2016",
-    "module": "commonjs",
+    "target": "ESNext",
+    "module": "ESNext",
```
kaladinlight, line 7: we should test and alpha release to confirm everything still works as intended when importing in web if you haven't. i feel like there were build issues surrounding the need for commonjs at some point for reasons I can't recall, but that was also half a decade ago... just sanity check

examples/sandbox/index.ts
```diff
@@ -1139,7 +1137,8 @@ $cosmosTx.on("click", async (e) => {
     return;
   }
   if (core.supportsCosmos(wallet)) {
-    const unsigned: core.Cosmos.StdTx = cosmosTransferTx;
+    // @ts-expect-error this wrong but at least now it's clear to us that it is
```
kaladinlight, line 1140: these ones would be straightforward to fix the mocks if we wanted to

tsconfig.json
gomesalexandre: I did try installing the verdaccio build in web, which worked fine for MM... Though definitely worth publishing an alpha instead of a regular release here so we can tackle changes if needed 🙏🏽

packages/hdwallet-ledger-webusb/src/transport.ts
gomesalexandre: 4193b21f

examples/sandbox/index.ts
kaladinlight: https://github.com/shapeshift/hdwallet/pull/681/commits/5bb93bb034193ccfa4e6cdcd7fadf78a0b0c4a2d

tsconfig.json
gomesalexandre: Full e2e paranoia test looking good:

- Native: https://jam.dev/c/9a441771-5018-4737-bc73-7dd3f0297432 (confirmed native unlock and wallet create is also happy).
  - noticing these warnings though seemingly unrelated? 
  image
- MM w/ snap: https://jam.dev/c/bcb3d215-0df7-435a-8fff-d47e5e828091
- Ledger: https://jam.dev/c/fff60d54-7fdd-4b1b-8a42-7e723975caf4
  - Unrelated to this PR, but noticing the ack isn't happy for RUNE since the "Pending for review" screen rugs the device detection, and it's an effective throw in swapper

kaladinlight (approved):

PR #684, opened 2024-09-21
fix: phantom ethAddress fast-follow

gomesalexandre:
Makes Phantom menu option go from this:

Screenshot 2024-09-21 at 11 45 21

To that:

Screenshot 2024-09-21 at 12 20 41


NeOMakinG (approved):
Before 
image

After
image

lgtm!

PR #689, opened 2024-10-03
feat: eip-6963 support

gomesalexandre:
a.k.a "The Big Short (2015)".

⚠️ Ready for conceptual review and early testing with its web companion, but **not** ready to merge until web is ready - few things to tackle there both from a technical and product perspective.
Most likely, nothing will change here in hdwallet-land as the changeset is extremely low and we shouldn't require anything more, though some smolish follow-ups could be made:

- Phantom leverages EIP provider
- Keplr EVM support is added through EIP provider

This PR brings EIP-6963 support through passing an rdns when instantiating metamask, and detecting the correct provider instead of `detectEthereumProvider`. That's it.
All the other changes here are effectively:

- removal of unused hdwallet-metamask (flag was off in web for a long time already and we weren't hitting its branch)
- removal of now unused hdwallet-xdefi
- removal of now unused hdwallet-coinbase
- removal of unused hdwallet-tallyho

See inline comments for the only actual change in this mist of red lines.


examples/sandbox/index.html
```diff
@@ -117,10 +117,7 @@ <h4>Select</h4>
       <button id="native">Pair Native</button>
       <button id="metaMask">Pair MetaMask</button>
       <button id="phantom">Pair Phantom</button>
-      <button id="xdefi">Pair XDEFI</button>
       <button id="keplr">Pair Keplr</button>
-      <button id="tallyHo">Pair Tally Ho</button>
-      <button id="coinbase">Pair Coinbase</button>
```
gomesalexandre, line 123: Removed across the board in this PR:

- Tally was dead for a long time already, and could ironically revive if Tally announced themselves as en EIP-6363 provider https://github.com/tahowallet/extension/issues/3725
- Coinbase doesn't need an explicit button anymore and is handled by the EIP detection
- Same as above for XDEFI

image

examples/sandbox/index.ts
```diff
@@ -9,16 +7,14 @@ import * as keepkeyWebUSB from "@shapeshiftoss/hdwallet-keepkey-webusb";
 import * as keplr from "@shapeshiftoss/hdwallet-keplr";
 import * as ledgerWebHID from "@shapeshiftoss/hdwallet-ledger-webhid";
 import * as ledgerWebUSB from "@shapeshiftoss/hdwallet-ledger-webusb";
-import * as metaMask from "@shapeshiftoss/hdwallet-metamask";
```
gomesalexandre, line 12: We weren't relying on hdwallet-metamask in web for a long, very long time already. Ditches the regular hdwallet-metamask since the multichain one superseeds it.

packages/hdwallet-metamask-shapeshift-multichain/src/adapter.ts
```diff
@@ -1,14 +1,13 @@
-import detectEthereumProvider from "@metamask/detect-provider";
```
gomesalexandre, line 1: We obviously don't need this anymore since we support MIPD

packages/hdwallet-metamask-shapeshift-multichain/src/adapter.ts
```diff
@@ -72,39 +71,33 @@ const METAMASK_IMPERSONATORS = [
...
-      silent: false,
-      timeout: 3000,
-    })) as providers.ExternalProvider | null;
-    if (!provider) {
-      const onboarding = new MetaMaskOnboarding();
-      onboarding.startOnboarding();
-      console.error("Please install MetaMask!");
-      throw new Error("MetaMask provider not found");
-    }
+    const maybeEip6963Provider = store.findProvider({ rdns: this.providerRdns });
```
gomesalexandre, line 94: Yep, that's all it takes. Didn't even need to pass it from web in the end (which didn't work, something something proxies?), which is actually good, in-place provider detection with rdns

gomesalexandre: Note naming: EIP-6963 structure here, with `provider` (an EIP-1193 compliant provider) and `info` (provider info) vs. 1193 below, the good ol' JS provider API https://eips.ethereum.org/EIPS/eip-1193

packages/hdwallet-metamask-shapeshift-multichain/package.json
```diff
@@ -21,7 +21,8 @@
     "@shapeshiftoss/metamask-snaps-adapter": "^1.0.10",
     "@shapeshiftoss/metamask-snaps-types": "^1.0.10",
     "eth-rpc-errors": "^4.0.3",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "mipd": "^0.0.7"
```
gomesalexandre, line 25: For ease of use instead of dealing with EIP-6963 reqs/event handlers ourselves, great lib from wevm as usual!

packages/hdwallet-trezor/src/trezor.ts
```diff
@@ -150,10 +150,6 @@ export class TrezorHDWalletInfo implements core.HDWalletInfo, core.BTCWalletInfo
     return Btc.btcGetAccountPaths(msg);
   }
 
-  public btcIsSameAccount(msg: Array<core.BTCAccountPath>): boolean {
```
kaladinlight, line 153: ? part of `BTCWalletInfo` interface. seems like we should still have this in some form or fashion

packages/hdwallet-core/src/bitcoin.ts
```diff
@@ -341,12 +341,6 @@ export interface BTCWalletInfo extends HDWalletInfo {
    */
   btcGetAccountPaths(msg: BTCGetAccountPaths): Array<BTCAccountPath>;
 
-  /**
-   * Does the device support spending from the combined accounts?
-   * The list is assumed to contain unique entries.
-   */
-  btcIsSameAccount(msg: Array<BTCAccountPath>): boolean;
```
kaladinlight, line 348: oh I see you removed it. major version bump change to interface? (no idea what the point of this function was for)

gomesalexandre: Yeah given the fact that we have removed quite a big of packages, we should definitely go with at the very least a minor vs. a patch as we usually do. Not sure about major here, something something about open-source. 

Looking at Trezor implementation, this is used to detect diff. purposes over same slip44 + accountId... which we can do, and actually do pretty easily in web already! 
image

packages/hdwallet-native/src/bitcoin.ts
```diff
@@ -35,6 +35,7 @@ type InputData = UtxoData | ScriptData | BchInputData;
 
 export function MixinNativeBTCWalletInfo<TBase extends core.Constructor<core.HDWalletInfo>>(Base: TBase) {
   // eslint-disable-next-line @typescript-eslint/no-shadow
+  // @ts-ignore tsc cache is drunk, btcIsSameAccount has been removed, TODO(gomes): rm ts-ignore
```
kaladinlight, line 38: dont' forget

packages/hdwallet-metamask-multichain/src/adapter.ts
```diff
@@ -0,0 +1,62 @@
+import * as core from "@shapeshiftoss/hdwallet-core";
+import { enableShapeShiftSnap, shapeShiftSnapInstalled } from "@shapeshiftoss/metamask-snaps-adapter";
+import { createStore } from "mipd";
+
+import { SNAP_ID } from "./common";
+import { MetaMaskShapeShiftMultiChainHDWallet } from "./shapeshift-multichain";
```
kaladinlight, line 6: we should update class name to match while we are at it

kaladinlight:
Main concern coming to mind with this is the fact that the other eip-6963 providers will now be "multichain", when in reality they aren't. all of our logic is dependent on the `supportsCoin` logic which wouldn't be accurate. This was why I was thinking a base eip-6963 implementation with the injected provider, and then everyone can extend that as they see fit (multichain, phantom, keplr, etc). I suppose that isn't entirely necessary for the phantoms/keplers since we will be instantiating their specific wallet in web, but for xdefi/coinbase, it seems applicable in that we would be using a wallet with multichain support when it doesn't have it (unless I am missing something).

gomesalexandre:
> Main concern coming to mind with this is the fact that the other eip-6963 providers will now be "multichain", when in reality they aren't. all of our logic is dependent on the `supportsCoin` logic which wouldn't be accurate. This was why I was thinking a base eip-6963 implementation with the injected (or detected) provider, and then everyone can extend that as they see fit (multichain, phantom, keplr, etc). I suppose that isn't entirely necessary for the phantoms/keplers since we will be instantiating their specific wallet in web, but for xdefi/coinbase, it seems applicable in that we would be using a wallet with multichain support when it doesn't have it (unless I am missing something).

As discussed with @kaladinlight, hdwallet improvements could be done as a follow-up (though hdwallet is perma-improvements land), but as it stands now, there is no change re: current flow, since we weren't using the regular `hdwallet-metamask` anymore for a long, long time already. See description for two follow-ups that would make sense effort-wise, but probably not worth doing any kind of MM snaps/pure EVM hdwallet split in hdwallet land as web is handling this at runtime anyway.

packages/hdwallet-native/src/bitcoin.ts
gomesalexandre: 27297629e92051d6442f056b0da56dd8a915a0f2 - keeping opened in case it comes back biting us to nuke again

packages/hdwallet-metamask-multichain/src/adapter.ts
gomesalexandre: a9fa9f5079c02d12f9800e9e182b88a5f8ffa141 (don't mind the diff, needed actual bimp for `integration` not to be borked... which I also added missing package.json entry for [there](https://github.com/shapeshift/hdwallet/commit/a9fa9f5079c02d12f9800e9e182b88a5f8ffa141#diff-cd67598767aad4b2e7c64086173ce0c83943deb401b2a81f8e3d835d8141cbc9R18), was shadowed previously)

examples/sandbox/package.json
```diff
@@ -21,7 +21,7 @@
     "@shapeshiftoss/hdwallet-ledger": "1.55.9",
     "@shapeshiftoss/hdwallet-ledger-webhid": "1.55.9",
     "@shapeshiftoss/hdwallet-ledger-webusb": "1.55.9",
-    "@shapeshiftoss/hdwallet-metamask": "1.55.9",
+    "@shapeshiftoss/hdwallet-metamask-multichain": "1.55.10-mipd.0",
```
gomesalexandre, line 24: Sandbox is a big mess - as far as I can understand from these stale refs, this doesn't use lerna but upstream (where `@shapeshiftoss/hdwallet-metamask-multichain` only exists as `1.55.10-mipd.0` alpha for now) , and trying to use 1.55.10 across the board here is a rug:
image

So looks like we're always one version behind in this package.json pretty much?

gomesalexandre: Actually shadowed *is* fine, that's what we want here not to always be one version behind and have to manually update: 7309f547
image

packages/hdwallet-metamask-multichain/src/adapter.ts
gomesalexandre: Actually: https://github.com/shapeshift/hdwallet/pull/689#discussion_r1787469657

examples/sandbox/package.json
gomesalexandre: Actually, shadowed *is* fine and that's the reason why things were borked in the first place with this package rename, and why we were always a version behind and had to manually update. 7309f547e9efc247c128c560c64ea1fe19dcf1a4

packages/hdwallet-core/package.json
```diff
@@ -26,5 +26,6 @@
   "devDependencies": {
     "@types/google-protobuf": "^3.15.1",
     "@types/lodash": "^4.14.168"
-  }
+  },
+  "gitHead": "27297629e92051d6442f056b0da56dd8a915a0f2"
```
kaladinlight, line 30: remove the `gitHead`s from all of the package jsons

examples/sandbox/package.json
```diff
@@ -21,7 +21,7 @@
     "@shapeshiftoss/hdwallet-ledger": "1.55.9",
```
kaladinlight, line 21: sandbox should ref 1.55.11 too

integration/src/wallets/metamask.ts
```diff
@@ -56,17 +60,17 @@ export function selfTest(get: () => core.HDWallet): void {
     expect(await wallet.ethSupportsNetwork()).toEqual(true);
   });
 
-  it("does not support BTC", async () => {
+  it("~~does not~~ does support BTC, because that's what real multichain is", async () => {
```
kaladinlight, line 63:
```suggestion
  it("supports Bitcoin", async () => {
```

integration/src/wallets/metamask.ts
```diff
@@ -56,17 +60,17 @@ export function selfTest(get: () => core.HDWallet): void {
...
+    expect(core.supportsBTC(wallet)).toBe(true);
   });
 
   it("does not support Native ShapeShift", async () => {
     if (!wallet) return;
     expect(wallet.ethSupportsNativeShapeShift()).toEqual(false);
   });
 
-  it("does not supports bip44 accounts", async () => {
+  it("does support bip44 accounts", async () => {
```
kaladinlight, line 73:
```suggestion
  it("does not support bip44 accounts", async () => {
```

integration/src/wallets/metamask.ts
```diff
@@ -83,23 +87,26 @@ export function selfTest(get: () => core.HDWallet): void {
 
   it("uses correct eth bip44 paths", () => {
     if (!wallet) return;
-    [0, 1, 3, 27].forEach((account) => {
+    // MM doesn't support multi-account
+    const accountNumbers = metamask.isMetaMask(wallet) ? [0] : [0, 1, 3, 27];
+    if (metamask.isMetaMask(wallet)) return;
```
kaladinlight, line 92: revert. no multi account wallet support, but can still get and describe paths as it stands.

integration/src/wallets/metamask.ts
```diff
@@ -83,23 +87,26 @@ export function selfTest(get: () => core.HDWallet): void {
...
+    accountNumbers.forEach((account) => {
       const paths = wallet.ethGetAccountPaths({
-        coin: "Ethereum",
+        coin: "ethereum",
```
kaladinlight, line 95: revert, pref to use `coin` as it appears in the `slip44Table`. can `toLowerCase` the `describePath` switch statement in `shapeshift-multichain` or just update to coin keys (uppercase)

packages/hdwallet-metamask-multichain/src/adapter.ts
```diff
@@ -0,0 +1,60 @@
...
+    return new MetaMaskAdapter(keyring, providerRdns);
+  }
+
+  public async initialize(): Promise<number> {
+    return Object.keys(this.keyring.wallets).length;
+  }
+
+  public async pairDevice(): Promise<MetaMaskMultiChainHDWallet | undefined> {
+    const maybeEip6963Provider = mipdstore.findProvider({ rdns: this.providerRdns });
+    if (!maybeEip6963Provider) throw new Error("EIP-6963 provider not found");
```
kaladinlight, line 27: we lost the onboarding logic here, do we want to trigger mm onboarding for `io.metamask` so things behave the same?

packages/hdwallet-metamask-multichain/src/mipdStore.ts
```diff
@@ -0,0 +1,3 @@
+import { createStore } from "mipd";
+
+export const mipdstore = createStore();
```
kaladinlight, line 3: save a file and just instantiate in the adapter?

packages/hdwallet-metamask-multichain/src/adapter.ts
gomesalexandre: Waiting on a reply from product here, intent is to use wc instead since deep linking is broken half/most of the time depending on platforms

packages/hdwallet-core/package.json
gomesalexandre: e8d0d107

integration/src/wallets/metamask.ts
gomesalexandre: e8d0d107

integration/src/wallets/metamask.ts
gomesalexandre: e8d0d107

packages/hdwallet-metamask-multichain/src/mipdStore.ts
gomesalexandre: e8d0d107

examples/sandbox/package.json
gomesalexandre: image

Doesn't look like we're able to do it in the same PR and are inherently one version ahead?

kaladinlight: we removed coinbase so that is why it isn't able to ref a local package

packages/hdwallet-metamask-multichain/src/adapter.ts
gomesalexandre: Disregard me - confirmed we want to keep the current behaviour *fully* as in, always display providers even if not installed (in the same order as currently), only appending the ones we don't currently have already

packages/hdwallet-metamask-shapeshift-multichain/src/adapter.ts
gomesalexandre: Actually we do

integration/src/wallets/metamask.ts
gomesalexandre: df02ebb

integration/src/wallets/metamask.ts
gomesalexandre: df02ebb2

examples/sandbox/package.json
gomesalexandre: 🧠 de2247e6

packages/hdwallet-metamask-multichain/src/shapeshift-multichain.ts
```diff
@@ -94,7 +96,7 @@ export class MetaMaskShapeShiftMultiChainHDWalletInfo implements core.HDWalletIn
       case "atom":
         return core.cosmosDescribePath(msg.path);
 
-      case "ethereum":
```
kaladinlight, line 97: really feels like we should do this to all of the coins in here to make sure they match the `slip44Table` (not that we really use this function anywhere, but for posterity)

integration/src/wallets/metamask.ts
```diff
@@ -83,7 +87,9 @@ export function selfTest(get: () => core.HDWallet): void {
 
   it("uses correct eth bip44 paths", () => {
     if (!wallet) return;
-    [0, 1, 3, 27].forEach((account) => {
+    // MM doesn't support multi-account
+    const accountNumbers = metamask.isMetaMask(wallet) ? [0] : [0, 1, 3, 27];
```
kaladinlight, line 91: this intergration test is only for metamask anyways, so ternary is kind of pointless. can just have `[0]` if we don't want to show that it still _is_ able to derive account paths even if not natively supported by the wallet itself

packages/hdwallet-metamask-multichain/src/adapter.ts
```diff
@@ -0,0 +1,68 @@
...
+    return new MetaMaskAdapter(keyring, providerRdns);
+  }
+
+  public async initialize(): Promise<number> {
+    return Object.keys(this.keyring.wallets).length;
+  }
+
+  public async pairDevice(): Promise<MetaMaskMultiChainHDWallet | undefined> {
+    const maybeEip6963Provider = mipdstore.findProvider({ rdns: this.providerRdns });
+    if (!maybeEip6963Provider) {
```
kaladinlight, line 30: we don't want to trigger the metamask onboarding if trying to pair a non MM wallet (xdefi/coinbase/tally/etc). otherwise just generic throw.
`if (!maybeEip6963Provider && isMetaMask) {`

packages/hdwallet-metamask-shapeshift-multichain/src/adapter.ts
kaladinlight: ?

packages/hdwallet-metamask-multichain/src/shapeshift-multichain.ts
gomesalexandre: All in for cleanup! e0e8f4e587c7b8143c9603066ad92ab08fe08762

integration/src/wallets/metamask.ts
gomesalexandre: e0e8f4e5

packages/hdwallet-metamask-multichain/src/adapter.ts
gomesalexandre: e0e8f4e5

kaladinlight: We want to still throw a generic error on `!maybeEip6963Provider`?

gomesalexandre: Absolutely - as discussed this wasn't really a concern previously (if we ever hit this, we could safely assume there *is* a provider since its rdns was originally detected in web) but since we're now adding a static list of mipd providers (regardless of installation), then we absolutely want to throw here in case we hit a static non-MetaMask rdns (Coinbase, XDEFI, Rabby) and it isn't installed: 1b4ad64b

examples/sandbox/package.json
```diff
@@ -12,25 +12,22 @@
...
-    "@shapeshiftoss/hdwallet-xdefi": "1.55.9",
+    "@shapeshiftoss/hdwallet-core": "1.55.10",
+    "@shapeshiftoss/hdwallet-keepkey": "1.55.10",
+    "@shapeshiftoss/hdwallet-keepkey-tcp": "1.55.10",
+    "@shapeshiftoss/hdwallet-keepkey-webusb": "1.55.10",
+    "@shapeshiftoss/hdwallet-keplr": "1.55.10",
+    "@shapeshiftoss/hdwallet-ledger": "1.55.10",
+    "@shapeshiftoss/hdwallet-ledger-webhid": "1.55.10",
+    "@shapeshiftoss/hdwallet-ledger-webusb": "1.55.10",
+    "@shapeshiftoss/hdwallet-metamask-multichain": "1.55.10-mipd.1",
```
kaladinlight, line 23: don't forget to change before merge

kaladinlight (approved):

PR #691, opened 2024-10-29
feat: only connect solana on pairDevice()

gomesalexandre:
### Description

Does what it says on the box, making connection some better for Solana accounts without an EVM address present.

The reason why EVM addy was used as a deviceId (and in `pairDevice()`) was because Solana support wasn't implemnented then, meaning we assumed EVM support in Phantom.
However, this is not always guaranteed to be true e.g with automagically discovered accounts that may not contain EVM accounts, and/or manually added accounts which may also not.

Fixed by not pairing EVM accounts on *instantiation*, which makes things some better, though the error is always present until we bring the same notion of runtime chains to Phantom as we do for Solana (which we probably should) 

### Issue

Tackles https://github.com/shapeshift/web/issues/7973

### Screenshots

https://jam.dev/c/be486369-d71c-44f0-831e-396caeac5c52


NeOMakinG (changes requested):
Work as intented with a solana account:
image

deviceId is `phantom:` prefixed which is the solana deviceId 

If I connect a ledger while only importing the ethereum account, it fails at the connect step:

image

image

NeOMakinG (approved):
re: private conversation

This is a compromise, if the solana account is not present it will fail, but it seems more a edge case than if the eth account is not here

Phantom is the only wallet supporting solana in our stack, we need solana to be reliable with it

PR #692, opened 2024-11-14
fix: adjust solana build tx function to work with programs

NeOMakinG:
- Making to and value optional in SolanaSignTx as when you use program you don't need a `to` and `value` as it's contained inside the instructions
- Also added the possibility to use `isSigner` and `isWritable` inside each instruction as multiple solana accounts can sign an instruction (delegating something etc), it's needed for jupiter swaps

### DIFF
--- a/packages/hdwallet-core/src/solana.ts
+++ b/packages/hdwallet-core/src/solana.ts
@@ -21,9 +21,9 @@ export interface SolanaTxInstruction {
 export interface SolanaSignTx {
   addressNList: BIP32Path;
   /** to is the destination pubkey for the transfer */
-  to: string;
+  to?: string;
   /** value is the amount to transfer in micro lamports*/
-  value: string;
+  value?: string;
   /** blockHash is used for expiry determination */
   blockHash: string;
   /** computeUnitLimit is the maximum number of compute units allowed to be consumed by the transaction */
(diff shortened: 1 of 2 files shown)


packages/hdwallet-core/src/solana.ts
```diff
@@ -21,9 +21,9 @@ export interface SolanaTxInstruction {
 export interface SolanaSignTx {
   addressNList: BIP32Path;
   /** to is the destination pubkey for the transfer */
-  to: string;
+  to?: string;
   /** value is the amount to transfer in micro lamports*/
-  value: string;
+  value?: string;
```
gomesalexandre, line 26: preferably-blocking: conceptually fine if not required to interact with PDAs, but not sure how I feel about this being in the main interface, introduces a lot of risks for all other Txs. 

Perhaps a program variant of this interface would make more sense instead?

NeOMakinG: Ah sorry I should have convert this to a draft, as seen with @kaladinlight we already zero out those parameters to comply with SPL usages, we might be able to do the same for any other program usage

I need to verify it works properly before closing this PR 🙏

NeOMakinG:
Confirming I could avoid this PR!

PR #701, opened 2024-12-16
feat: keepkey taproot support

gomesalexandre:
### Description

Update: it now does! ~~"ish", as in we're building a correct payload that KeepKey is unable to understand because of firmware itself not supporting P2TR outputs.~~ 
This is more than likely related to https://github.com/keepkey/device-protocol being the source-of-truth for messaging on the device, and said library not being up-to-date with latest `@trezor/protobuf` (which `@keepkey/device-protocol` should be a superset of, and by extension, the device should understand the latest spec of).

### Testing

Able to send P2TR outputs with the usual verdaccio business in https://github.com/shapeshift/web/pull/8951.

### Issue

Relates to https://github.com/shapeshift/web/issues/8277


gomesalexandre:
Note: this may be required for https://github.com/shapeshift/web/pull/8951 to work. If so, don't forget to lerna bump here.

gomesalexandre:
Closing as not required in the end 🎉

PR #706, opened 2025-01-08
chore: keystore suggestions

kaladinlight:
- more generalized keystore handling (thorswap + mew)
- we _shouldn't_ need the register transformer, but haven't tested officially
- leverage `blake2bHex` over our own wrapper function


gomesalexandre (approved):
Changes applied with the addition of 7c684a29153987ca838abaa57fa999546ccce3ab

PR #707, opened 2025-01-23
fix: make wcv2 Tx broadcast spec-compliant

gomesalexandre:
### Description

After way too many tries across different wallets and sanity checking against official examples, it looks like the reason why *some* wallets were failing signing was the `chainId` param

See https://github.com/reown-com/web-examples/blob/16d22815c0100bc4525afc25349d67d9e93f2b47/advanced/dapps/react-dapp-v2-with-ethers/src/helpers/tx.ts#L30 in official examples, where `chainId` 
is missing, as it should.

MM is perfectly happy with this param, but e.g 1inch wallet will fail if present.

And it makes absolute sense some wallets would not be happy with it, as it is *not* part of the official JSON-RPC spec: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_sendtransaction
MM flavour (with EIP-1559 support) also doesn't explicitly support it, but they may well just not be so strict WRT validation of additional params: https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_sendTransaction/

Removed `chainId` to keep things spec-compliant, and both 1inch and MM are now happy:

- 1inch: https://jam.dev/c/bd9cdd52-64c4-4726-b7a2-9fbcb5dc592e
- MM: https://jam.dev/c/ff9e5ae4-9e40-44bd-a571-e4d366ab190a

### Testing

- Test with 1inch Mobile and MM mobile through wc 
- Confirm *Tx* (not message) signing is happy e.g approvals or Tx broadcast

### Issue

- relates to https://github.com/shapeshift/web/issues/8645


NeOMakinG (approved):
image
1inch

image

MM

Looks good!

PR #712, opened 2025-03-25
fix: add ethers dependency to hdwallet-ledger

NeOMakinG:
It's needed if we use hdwallet-ledger outside of the vite patch with ethers 6 and ethers 5


packages/hdwallet-ledger/package.json
```diff
@@ -19,14 +19,15 @@
...
-    "@shapeshiftoss/hdwallet-core": "1.59.2",
+    "@shapeshiftoss/hdwallet-core": "1.59.3",
     "@solana/web3.js": "1.95.8",
     "base64-js": "^1.5.1",
     "bchaddrjs": "^0.4.4",
     "bitcoinjs-message": "^2.0.0",
     "bs58check": "2.1.2",
     "ethereumjs-tx": "1.3.7",
     "ethereumjs-util": "^6.1.0",
+    "ethers": "5.7.2",
```
0xApotheosis, line 30: 🧠

NeOMakinG: Yeah I was about to say that this is the actual diff

0xApotheosis (approved):
Great 🕵️ work neo, thanks for diving into this ser. I know it was a pain.

0xApotheosis:
Will confirm CI passes and then merge.

PR #718, opened 2025-04-24
feat: mayachain support native wallet

kaladinlight:


gomesalexandre (approved):
Tested in https://github.com/shapeshift/web/pull/9403#pullrequestreview-2795161556 but stamping for sanity here after code pass, pretty straightforward copypasta from THOR!

PR #723, opened 2025-06-16
feat: phantom turn _supportsBase flag

gomesalexandre:
### Description

Looks like we missed Base go-live on Phantom, this fixes it:

image

### Issue 

N/A

### Testing

N/A, only toggles chain support flag, testing TBD in web PR once published


0xApotheosis (approved):
Nice catch!

PR #724, opened 2025-06-24
feat: viem hex types for `ETHSignTx`

gomesalexandre:
See the comments on top of each of those. What's even better than comments? Type-safety baby :fridaydog:

Ensure that hexlified things are properly typed as such, so we have type-safety in web and ultimately can avoid things like https://github.com/shapeshift/web/issues/9772

Web PR fren:  https://github.com/shapeshift/web/pull/9774


packages/hdwallet-core/src/ethereum.ts
```diff
@@ -12,6 +12,9 @@ export interface SerializedEthereumRpcError {
   stack?: string;
 }
 
+type Address = `0x${string}`;
+type Hex = `0x${string}`;
```
kaladinlight, line 16: probably worth exporting if needed.

gomesalexandre: isn't for the time being in hdwallet-core, so probably not worth it?

kaladinlight: more from a package consumption point of view. not a huge deal since we will be using the comparable viem type in web

packages/hdwallet-core/src/ethereum.ts
```diff
@@ -46,17 +49,17 @@ export type ETHSignTx = {
   /** bip32 path to sign the transaction from */
   addressNList: BIP32Path;
   /** big-endian hex, prefixed with '0x' */
-  nonce: string;
+  nonce: Hex;
```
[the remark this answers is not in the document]
gomesalexandre: Welp, it's exported now!

kaladinlight (dismissed):
looks good, let's bump and send!

kaladinlight (approved):

